Issue #4, opened 2018-06-11
Add support for QAbstractItemModel base classes


tones111:
In order to use a tree view the underlying model must inherit from QAbstractItemModel which is currently not supported.

I have started a naive attempt to add support, but am struggling getting the generated code to compile.  One challenge is that the index and parent methods return a QModelIndex which I believe is causing the build to fail with

cargo:warning=.../target/debug/build/qmetaobject-90f6bc2d732d83be/out/rust_cpp/cpp_closures.cpp: In instantiation of ‘union rustcpp::return_helper<QModelIndex>::container::<unnamed>’:
cargo:warning=.../target/debug/build/qmetaobject-90f6bc2d732d83be/out/rust_cpp/cpp_closures.cpp:20:9:   required from ‘struct rustcpp::return_helper<QModelIndex>::container’
cargo:warning=.../target/debug/build/qmetaobject-90f6bc2d732d83be/out/rust_cpp/cpp_closures.cpp:27:45:   required from here
.../target/debug/build/qmetaobject-90f6bc2d732d83be/out/rust_cpp/cpp_closures.cpp:20:19: error: ‘rustcpp::return_helper<T>::container::<unnamed union>::memory’ has incomplete type
cargo:warning=         union { T memory; };
cargo:warning=                   ^~~~~~

I'm not sure what/where support for returning that type needs to go.  Any suggestions?

Thanks

ogoffart:
Difficult to say without seeing the code.
Maybe some includes are missing? 
Surely the documentation can be improved.

ogoffart:
Was fixed with #5

PR #5, opened 2018-06-11
Read-only QAbstractItemModel support

tones111:
As discussed in #4 

This is a very early work-in-progress to add the ability to specify QAbstractItemModel as a base class.  The generated code fails to compile.

The approach thus far has been to copy the listmodel code and make adjustments.  The idea is to get a read-only model working and then add additional methods.

### DIFF
--- a/qmetaobject/src/itemmodel.rs
+++ b/qmetaobject/src/itemmodel.rs
@@ -1,0 +1,151 @@
+/* Copyright (C) 2018 Olivier Goffart <ogoffart@woboq.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+associated documentation files (the "Software"), to deal in the Software without restriction,
+including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial
+portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES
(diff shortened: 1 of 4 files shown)


ogoffart:
So the error is because #include <QtCore/QAbstractItemModel> is not yet included at the time of declaration of the extern "C" function for the rust! macro.
This is a bit confusing and will need a better documentation on the use of the rust! macro.

The way to solve it is like this:

```rust
cpp!{{
 #include <qmetaobject_rust.hpp>
 #include <QtCore/QAbstractItemModel>
}}

cpp! {{
￼ struct Rust_QAbstractItemModel : RustObject<QAbstractItemModel> {
 ...
```

By splitting into two cpp! block, the extern "C" declaration will be put between these blocks when the types are already defined.

￼

tones111:
The next hurdle is that classes derived from QAbstractItemModel need to implement the [index method](https://doc.qt.io/qt-5/qabstractitemmodel.html#index) which returns a [QModelIndex](https://doc.qt.io/qt-5/qmodelindex.html#details).  Unfortunately the QModelIndex docs state that QAbstractItemModel::createIndex must be used for construction and both method signatures are protected.

Attempting to compile now results in an error.  Any ideas or suggestions on how to create the QModelIndex that needs to be returned from some of the functions?

error: ‘QModelIndex QAbstractItemModel::createIndex(int, int, void*) const’ is protected within this context
cargo:warning=             return obj ? obj->createIndex(row, column) : QModelIndex();
cargo:warning=                                                      ^
cargo:warning=In file included from /usr/include/qt/QtCore/QAbstractItemModel:1,
cargo:warning=                 from /home/paul/src/gitr/qmetaobject-rs/qmetaobject/src/itemmodel.rs:69:
cargo:warning=/usr/include/qt/QtCore/qabstractitemmodel.h:377:20: note: declared protected here
cargo:warning= inline QModelIndex QAbstractItemModel::createIndex(int arow, int acolumn, void *adata) const
cargo:warning=                    ^~~~~~~~~~~~~~~~~~
exit code: 1

ogoffart:
Add
```cpp
using QAbstractItemModel::createIndex;
```
in the public part of the Rust_QAbstractItemModel class to expose it as public

tones111:
I should have noticed that from the list model code. thanks.

Looks like read-only support works now.  I have a simple driver that creates an instance and provides a list to qml.  This seems like a good point to get this infrastructure pushed in and I'll start looking at adding more methods.  Thanks for your help.

ogoffart:
I merged your contribution. Thanks for it.
Sorry for the delay, i kind of forgot about it.

If you want to contriute more, i can give you contributor access to this repository. Would you be interrested?

PR #10, opened 2018-09-18
Use full include paths to fix compilation

vandenoever:


ogoffart:
Thanks.

Originally these files were coming from Qt and I put them as it,  but i guess it is ok to change them a bit.

BTW, If you want to contribute more, i can give you write access to the repository.

PR #16, opened 2018-10-20
Specify -qt5 in qmake command

Derriick:
Generally, the default version of Qt used by qtchooser (and therefore qmake) is Qt4


ogoffart:
Thanks for your patch!

However that's not that simple. Because this fails when qtchooser is not installed (as it is the case on my archlinux, or on windows, see the appveyor error.)

Maybe you can try your luck with the `QT_SELECT` env variable?

Derriick:
OK I used to think that qtchooser was installed with qmake package of any distribution.
The QT_SELECT env variable works well, thanks for support :)

PR #17, opened 2018-10-20
Add link to cpp! macro example code

naufraghi:


ogoffart:
Thanks for the patch!

PR #18, opened 2018-10-26
Add ability to initialize the QtWebEngine, gated behind a feature

FuegoFro:
This adds a new `webengine` feature to the crate which, when enabled,
adds the ability to initialize the `QtWebEngine` and links against the
corresponding libraries.

### DIFF
--- a/qmetaobject/Cargo.toml
+++ b/qmetaobject/Cargo.toml
@@ -9,6 +9,10 @@ license = "MIT"
 keywords = ["Qt", "QML", "QMetaObject",]
 repository = "https://github.com/woboq/qmetaobject-rs"
 
+[features]
+default = []
+webengine = []
+
 [dependencies]
 qmetaobject_impl = { path = "../qmetaobject_impl", version = "=0.0.3"}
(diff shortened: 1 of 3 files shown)


FuegoFro:
I'm somewhat new to Rust, so let me know if there's better ways to do any of these changes! Thanks!

ogoffart:
I'm new to rust myself.

The problem here is that the cpp_build crate does not know about the cfg feature gate and will generate the code no matter what.
I don't really have a solution to this problem.  Always depending on QtWebegine is maybe a bit overkill. Maybe this could be in a different crate, but that's a pity.  FInd an option so cpp_build understands the cfg may be another way.

If I understand correctly, you need to call this to use a web engine item in a qml file?

Also, it is missing documentation, of both the feature and the new API.

FuegoFro:
Hey, sorry for leaving this open for so long! I haven't had time to work on this in a while, so I'm going to go ahead and close this out. If I come back to it, I'll make sure to fix the issue you pointed out (thank you for that!).

PR #22, opened 2019-01-06
Add API wrapper for QQuickStyle::setStyle

timsueberkrueb:
I'm new to `qmetaobject-rs` so please correct me if this should be done differently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
     - sudo apt-get update
 
 install:
-    - sudo apt-get install -y -qq qt${QT_VERSION}base qt${QT_VERSION}declarative g++-7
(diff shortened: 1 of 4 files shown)


timsueberkrueb:
Okay, will do, thanks for the feedback!

timsueberkrueb:
Which one would you prefer? Using the envvar or adding it to `QmlEngine`? I personally would prefer the `QQuickStyle` version as it mirrors the Qt api => therefore less surprising.

ogoffart:
You are right, mimicking the QQuickStyle API is probably the best

timsueberkrueb:
Done. I'm using `QString` because it matches the Qt API most closely. Is this okay or should I use `&str`?

timsueberkrueb:
Unrelated: I think it might also make sense to use a binding generator to offer a more complete binding set on top of qmetaobject-rs, perhaps as a different crate. What do you think?

Issue #29, opened 2019-02-26
Assigning/binding custom property types within qml.


Pfeil:
(I created a [minimal example][repo] which may be helpful to take a closer look. This uses the normal qmetaobject version from crates.io. All code snippets here are excerpts from this repo.)

[repo]: https://github.com/Pfeil/qml_custom_property_minimal

So I created two types inheriting from QQuickItem. One should be a property from the other. I've done this via a RefCell as you told me in #26 .
```
#[derive(Default, QObject)]
struct Slice {
    base: qt_base_class!(trait QQuickItem),
}

#[derive(Default, QObject)]
struct Pie {
    base: qt_base_class!(trait QQuickItem),
    slice: qt_property!(RefCell<Slice>; /*READ get_slice WRITE set_slice*/),
}

impl Pie {
    //fn get_slice(&self) -> RefCell<Slice> {
    //    self.slice // how?
    //}
    //fn set_slice(&mut self, s: RefCell<Slice>) {
    //    self.slice = s;
    //}
}
impl QQuickItem for Slice {}
impl QQuickItem for Pie {}
```

While this compiles, QML does throw runtime errors. This is the QML file (without imports):

```
ApplicationWindow {
    Slice {id: s}  // works
    Pie {}    // works

    // does not work:
    Pie {
        slice: s
    }

    // does not work:
    Pie {
        slice: Slice {}
    }
}
```

So, the following things do currently not work:
1. binding the Slice to the Pie,
2. binding the Slice to the Pie using an identifier,
3. implementing a getter (and the setter may be garbage?).

From the tutorial and other qt docs I found, I was not able to derive whats wrong/missing, and in this area the examples and documentation of qmetaobject is getting thin. I'd love to work on the docs as soon as I finished what I began (and understand it ;) ).

ogoffart:
Thanks for reporting. I believe (not 100%sure) that the problem is that the Slice is registered twice as different QMetaType.

The first time it is register with you calling `qml_register_type` which you need to do to be able to use `Slice {}`
The second time, it is the `impl<T> PropertyType for ::std::cell::RefCell<T>` which registers again in `register_metatype_qobject`

Ideally, registering twice with the same name should merge both type internally within Qt. But apparently this is not the case.  I'll try to have a look next week.

Pfeil:
[...]

PR #41, opened 2019-07-10
Implement PropertyType for QPointer<T> (proposed in #29)

Guiguiprim:
It is useful to be able to have a settable property that is a QObject.


ogoffart:
Thanks!

PR #43, opened 2019-07-17
Enum property

Guiguiprim:
Add property support for enums.

I'm having a hard time finding the good way of registering correctly the enum.
I'm going to keep searching, but in the mid-times if you have any input on what I'm doing, they would be welcomed. 

### Registering
I introduced a function `register_metatype_enum<T: QEnum>()` but:
- The implementation I tried based on what was done for `QMetaType` doesn't work
```QML
  obj.prop: Enum.Value // cannot assign int to MyEnum
```
- As a test I returned the meta-type id of `i32`, that make it half work (by making the QML think it's only an integer...)
 ```QML
  obj.prop: Enum.Value // works
  obj.prop: 958 // doesn't generate an error as it should
```

### QMetaObject data
I'm also not really sure about the way I did some meta-object generation changes:
- Adding the enum flag to the properties
- Adding the related meta-objects into a meta-object (not absolutely sure I need them actually but the QML stack seems to use it for enumerators)

For both of them I needed to edit the data stored inside QMetaObject during the lazy static initialization 

### Contributing
On a more general aspect:
- Are you using `rustfmt` / `cargo fmt` ? If yes in which version ? I'm using format by default with rust 1.36 and a lot of the code-base is affected.
- I forgot to check for it in my previous contributions but, are you using clippy for a warnings free code ?

### CODE REVIEW

**Comment by ogoffart** [comment]
> Are you using rustfmt / cargo fmt ? 

I have been using it at some point. But i really did not like how it formated the cpp! macro so i have stopped using it.
In other words: I am not using cargo fmt.
But I guess I could just run it once.

**Comment by ogoffart** [comment]
I guess the best would be that QMetaType was automatically there for type that inherit from QEnum.

Something like `impl<T: QEnum> QMetaType for T`  would be nice but can't work since it conflict with the one for QGadget.

Since both QEnum and QGadget are implemented by the macro, i guess the best would be to implement QMetaType for these types within the procedural macro.

For QEnum, we can even add CONVERSION_TO_STRING and CONVERSION_FROM_STRING function.  And also when registering, we should pass the C++ QMetaType::IsEnumeration flag

**Comment by ogoffart** [comment]
> `obj.prop: Enum.Value // cannot assign int to MyEnum`

Adding the `QMetaType::IsEnumeration` flag while registering the C++ QMetaType should fix that problem.

> `obj.prop: 958 // doesn't generate an error as it should`

I believe QML does not generate an error for that also for enum upstream.
So that's not possible to fix.

> Adding the enum flag to the properties 

Looks good to me,  but I think i'd rather add the
`const IS_ENUM: bool = true;` to QMetaType then  (because i would rather not reimplement PropertyType from the macro.)
But that's a bit annoying that it is in the public API.

I am not sure if you really need to add the enum flag and extra properties.
This was done from before Qt has the ability to register the QMetaObject* of enums to QMetaType.  So it may work without it, as long as you add the `QMetaType::IsEnumeration` flag when registering the enum.


￼

**Comment by ogoffart** [comment]
Note that when re-implementing `QMetaType`, the implementation of `from_qvariant` should be checking that it is in the correct range

**Comment by ogoffart** [comment]
Actually you do not need to specify CONVERSION_TO_STRING / CONVERSION_FROM_STRING, as this is builtin if the QMetaType has a metaobject and is the flag QMetaType::IsEnumeration

**Comment by Guiguiprim** [comment]
I did check in Qt/cpp if you have an enum property (at least `enum class`) and try in QML to assign a integer value that is not in the enumerator you get an error (with log)

**Comment by Guiguiprim** [comment]
> QMetaType::IsEnumeration
I did see it at some point when reading Qt code and forgot about it...

**Comment by Guiguiprim** [comment]
Yep I was just missing QMetaType::IsEnumeration !
I removed all the stuff in meta-object generation (flag and related), and it works.

Yes, I did try `impl<T: QEnum> QMetaType for T` but as you said it conflict with `impl<T: QGadget> QMetaType for T`
I still implement `PropertyType` instead of `QMetaType`. Should I change it ?

**Comment by Guiguiprim** [comment]
As an improvement, I was also thinking of adding an attributes to QEnum derive to be able to give the default value (Default can't be derived for an enum).

**Comment by ogoffart** [comment]
> I still implement PropertyType instead of QMetaType. Should I change it ?

Yes, it makes more sense as QMetaType is also used for conversion to/from QVariant, which makes sense.

**Comment by Guiguiprim** [comment]
This time it should be good.
`#derive(QEnum)]``implements `QMetaType` !

Fun fact, if you have two `cpp!(...)` with identical content, you get a double definition c++ compilation error.

**Comment by ogoffart** [comment]
> Fun fact, if you have two cpp!(...) with identical content, you get a double definition c++ compilation error.

Should be fixed with the version 0.5.3 of the cpp crate.

File: qmetaobject/src/qmetatype.rs [hunk 1]
```diff
@@ -144,6 +147,45 @@ fn register_metatype_qobject<T: QObject>() -> i32 {
...
+}
+
+pub fn enum_from_qvariant<T: QEnum + QMetaType>(mut variant: QVariant) -> Option<T> {
+    let id: i32 = T::id();
+    let var_ptr = &mut variant as *mut QVariant;
+    let ptr = qvariant_internal_ptr(var_ptr, id);
+    if ptr.is_null() {
+        None
+    } else {
+        let raw = unsafe{ *(ptr as *const usize) };
```
ogoffart: I'm not sure if `usize` is right there.

ogoffart: It should actually be the underlying type  (the one in the repr)  You could add an associated type to the QEnum trait for that.

qmetaobject/src/qmetatype.rs
```diff
@@ -144,6 +147,45 @@ fn register_metatype_qobject<T: QObject>() -> i32 {
...
+    name: *const std::os::raw::c_char,
+) -> i32 {
+    register_metatype_common::<T>(name, T::static_meta_object(), true)
+}
+
+pub fn enum_to_qvariant<T: QEnum + QMetaType>(e: &T) -> QVariant {
+    let id: i32 = T::id();
+    let raw = e.to_raw_value() as usize;
+    cpp!(unsafe [id as "int", raw as "const void*"] -> QVariant as "QVariant" {
+        return QVariant(id, raw);
```
ogoffart: you are supposed to pass a pointer to the value here,  not the value itself.

qmetaobject/src/qmetatype.rs
Guiguiprim: `trait QEnum` as conversion to and from `u32` which is the bigger supported integer type (as it is stored in a array of u32 in the `QMetaObject` data).

But here I convert `u32` to `usize` as to store it in QVariant we need to convert it to `*const c_void`. We get an error when doing so from `u32`:
```
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> src/qmetatype.rs:159:31
    |
159 |     cpp!(unsafe [id as "int", raw as "const void*"] -> QVariant as "QVariant" {
    |                               ^^^
    |
    = note: source type: `u32` (32 bits)
    = note: target type: `[u8; 8]` (64 bits)
```
This is on the to_variant side. So I take the same way out.

ogoffart: Because it should be:
```
   cpp!(unsafe [id as "int", raw as "int"] -> QVariant as "QVariant" {
￼	        return QVariant(id, &raw);
```

I'm not sure why the tests are passing so there might be something wrong.

ogoffart: Well, even that would not be correct because raw should not be int, but the underlying type.  Because QVariant knows the size of the enum, as it was constructed.
So either you need to always give a size of 4 byte.  Or you need to use the proper underlying type here.

qmetaobject/tests/tests.rs
```diff
@@ -606,47 +606,172 @@ fn panic_when_moved_setter() {
...
+        0,
+        CStr::from_bytes_with_nul(b"OtherEnum\0").unwrap(),
+    );
+    qml_register_type::<MyEnumObject>(
+        lib,
+        1,
+        0,
+        CStr::from_bytes_with_nul(b"EnumObject\0").unwrap(),
+    );
+    let my_obj = MyEnumObject::default();
```
ogoffart: unused variable

Guiguiprim: Yep, I just removed it.

ogoffart:
You are actually not testing the from_qvariant,  and only to_qvariant with a '0' enum.

Maybe you can add a slot taking an enum as a parameter, and also returning one,  and initializing the enum to some other value.

Using special representation like u16  would also help to have more coverage

Guiguiprim:
So you are suggesting something like:
```Rust
trait QEnum {
  type Repr;
  fn to_raw(&self) -> Repr;
  fn from_raw(raw: Repr) -> Option<Self>;
}
```

Guiguiprim:
Your right the tests coverage is pore. I will try to improve it.

qmetaobject/src/qmetatype.rs
```diff
@@ -71,12 +72,14 @@ fn register_metatype_common<T: QMetaType>(
...
                 creator_fn as "QMetaType::Creator", destructor_fn as "QMetaType::Destructor",
-                constructor_fn as "QMetaType::Constructor", gadget_metaobject as "const QMetaObject*"] -> i32 as "int" {
-            QMetaType::TypeFlags extraFlag(gadget_metaobject ? QMetaType::IsGadget : 0);
-            return QMetaType::registerType(gadget_metaobject ? gadget_metaobject->className() : name, deleter_fn, creator_fn, destructor_fn,
+                constructor_fn as "QMetaType::Constructor", metaobject as "const QMetaObject*",
+                is_enum as "bool"] -> i32 as "int" {
+            auto type_flag = is_enum ? QMetaType::IsEnumeration : QMetaType::IsGadget;
+            QMetaType::TypeFlags extraFlag(metaobject ? type_flag : 0);
+            return QMetaType::registerType(metaobject ? metaobject->className() : name, deleter_fn, creator_fn, destructor_fn,
                 constructor_fn, size,
```
ogoffart: here, `size` should match the  the actual size behind the pointer which is passed to the QVarient.
So if you're gonna always use u32,  it should always be sizeof(u32)  for enums

ogoffart:
(If you're gonna add many more tests, you could consider adding a new file like tests/enum.rs)

qmetaobject/src/qmetatype.rs
Guiguiprim: Seems legit.
So either I stick to `u32` and fix it here, or I add the representation to the `QEnum`.
Got it, make sense.

Guiguiprim:
Not there yet.

- I made the changes to add `type Repr` in `QEnum`
- I made some better tests

Some tests fail and for now I'm going nowhere:
- The `Peach` assignment fail, enums.rs:135 logs an empty string instead of `Peach`
- `price_level_of()` doesn't return correctly into QML (when adding a log of the return in QML, I get random values printed...)

qmetaobject/tests/enums.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use std::ffi::CStr;
+
+mod common;
+use common::*;
+
+#[derive(QEnum, Clone, Copy, Debug)]
+#[repr(i8)]
+enum PriceLevel {
+    Cheap = -23,
```
ogoffart: that's indeed quite cheap :-P

qmetaobject/src/qmetatype.rs
```diff
@@ -144,6 +147,46 @@ fn register_metatype_qobject<T: QObject>() -> i32 {
     }
 }
 
+pub fn register_metatype_qenum<T: QEnum + QMetaType>(
```
ogoffart: You should add #[doc(hidden)]  To all these internal things that are used only from the macro.

ogoffart:
Have you made progress?

Now I realize that it is not valid to just cast from the QVariant's data to the rust enum,  because the Qt code might have put any integer in there.
So maybe we need something like a `QMetaType::verify(x: *const Self) -> bool` that check that the value is indeed approperiate.

For other types than enum this is ok to cast since QVariant ensure that the type is correct.  But when the enum flag is set, it just can convert from any int.

Guiguiprim:
I did not made any progress on it. (I'm not working on it lately)

The way I went about it, is to check the value:
- either it is a valid value and I set the value to it
- or I set the value to `default()`
This is done using `QEnum::from_raw(raw: Self::Repr) -> Option<Self>`

ratijas:
I found myself in need of this feature, so I should probably take a closer look, and get back to contributing.

*\*sigh\** that day job planning though…

Guiguiprim:
I forgot about this and I'm not working on it anymore.
Should I close it or leave it open if anyone want to take it over ?

PR #46, opened 2019-07-24
WIP: Use 'dyn' since trait objects without an explicit 'dyn' are deprecated

Atul9:


ogoffart:
I think it's fine to leave out the `dyn` when doing  `qt_base!(trait QObject)`   otherwise the proc macro would need to be adapted.

Do you see warnings for that?

Atul9:
yes, I saw warnings in this build https://travis-ci.org/woboq/qmetaobject-rs/jobs/561092338 and hence I added dyn to `qt_base!(trait QObject)` -

Atul9:
@ogoffart yes, I saw warnings in this build https://travis-ci.org/woboq/qmetaobject-rs/jobs/561092338 and hence I added dyn to qt_base!(trait QObject)

Atul9:
@ogoffart I removed `dyn` from `qt_base!(trait QObject)` & updated the PR but the build fails https://travis-ci.org/woboq/qmetaobject-rs/jobs/563174204, 
```Failed to fetch http://ppa.launchpad.net/beineri/opt-qt-5.10.1-trusty/ubuntu/dists/xenial/main/binary-amd64/Packages  404  Not Found```
I force pushed the branch to check if the package will be fetched in the next build but that build failed with the same error. what do you suggest?

ogoffart:
Sorry for not coming back to you earlier. But now there is already `dyn` added in master
Thanks for your commit anyway.

It appear that there is indeed a problem in he travis  about fetching Qt

PR #49, opened 2019-08-20
Fix `dyn Trait` warnings for rust 1.37

Defaultldentity:
I could only build `todos` example though; not sure what to do with the others.


ogoffart:
Thank you.

PR #54, opened 2019-10-20
Add implementations of From<f32> and From<f64> to QVariant

agentsim:


ogoffart:
Thanks!

(A test would have been nice)

PR #56, opened 2019-10-21
Add QDate, QTime, QDateTime and From<...> Implementations for QVariant

Ace4896:
Needed wrappers for Qt's various date/time classes for a side-project I'm currently working on, so I've added these. This also adds a new optional feature `chrono_qdatetime` to the crate which enables conversion from chrono's date/time types to their respective Qt wrappers.

Still a WIP since I'm not sure about:

- How to deal with the API for QDateTime; this is timezone aware like chrono's DateTime and has various constructors. Should I also try to wrap the Qt::TimeSpec enum and the QTimeZone class? 
- Where to write tests - in the `tests` directory or in `src/qttypes.rs`?

I don't usually write C++, so let me know if I've done anything glaringly wrong in the C++ wrapper functions.

P.S. I noticed there was no .gitignore; I've added one in this PR which ignores `target` and `Cargo.lock`.


ogoffart:
Looks good so far. Thanks for the contribution.

For the tests, it is up to you, but I'd put simple tests "inline" and more complex tests in the tests directory. Also consider documentation example tests.

You will also need to implement the qmetatype trait in qmetatype.rs I believe these are built ins, so you can use the macro

As for the API, I would keep it simple. As long as it can be extended should we need it. So it is ok to omit timezone support and add that later.

Ace4896:
Alright, so I've left out the timezone things for QDateTime, and assumed local/system timezone when constructing one using both QDate and QTime.

I also changed the AppVeyor and Travis CI configurations so that they test all the features in the crate as well (because of the new optional feature I added).

The tests I added are just inline ones; didn't think they needed to go into the other directory. I couldn't figure out why the QTime tests were failing on my laptop; for some reason, the chrono variant of the test was passing, but the normal one wasn't (despite both of them using the same functions). I'll see what the CI run says when it finishes.

ogoffart:
I fixed the bug and pushed the work so far.
It turned out the problem was just that creating a QTime with 1000 ms was creating an invalid QTime.

Thanks for the contribution, feel free to add more functions if you need to.

Ace4896:
Ah, good spot, I didn't think that the test values I put would be wrong (forgot that 1000ms = 1s 😅)

I'll be sure to contribute more if I find anything else I need.

PR #57, opened 2019-10-25
Create a wrapper around QQmlComponent

agentsim:
Support setData and loadUrl methods
Fix SignalCppRepresentation to support classes other than QObject

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -278,67 +278,18 @@ impl QmlComponent {
         })}
     }
 
+    /// Performs QQmlComponent::status
+    pub fn status(&self) -> ComponentStatus {
+        unsafe { cpp!([self as "QQmlComponentHolder*"] -> ComponentStatus as "QQmlComponent::Status" {
+            return self->component->status();
+        })}
+    }
+
     /// See Qt documentation for QQmlComponent::statusChanged
     pub fn status_changed_signal() -> CppSignal<fn(status: ComponentStatus)> {
         unsafe { CppSignal::new(cpp!([] -> SignalCppRepresentation as "SignalCppRepresentation"  {
             return &QQmlComponent::statusChanged;
         }))}
-    }
-}
-
-#[cfg(test)]
-mod qqmlcomponent_tests {
-    use crate::*;
-    use std::cell::RefCell;
-    use std::rc::Rc;
-
-    #[test]
-    fn create_component() {
-        let qml_text = "import QtQuick 2.0\nItem{}".to_owned();
-
-        let engine = QmlEngine::new();
-        let mut component = QmlComponent::new(&engine);
-
-        component.set_data(qml_text.into());
-
-        let obj = component.create();
-
-        assert!(!obj.is_null());
-    }
-
-    #[test]
-    fn component_status_changed() {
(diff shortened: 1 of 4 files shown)


qmetaobject/src/qtdeclarative.rs
```diff
@@ -201,6 +202,106 @@ impl Default for QQuickView {
     }
 }
 
+#[repr(u32)]
+#[derive(Clone, Copy)]
+pub enum CompilationMode {
```
ogoffart: Please add documentation

qmetaobject/src/qtdeclarative.rs
```diff
@@ -201,6 +202,106 @@ impl Default for QQuickView {
...
+    PreferSynchronous,
+    Asynchronous,
+}
+
+#[repr(u32)]
+#[derive(Clone, Copy)]
+pub enum ComponentStatus {
```
ogoffart: Please add documentation

qmetaobject/src/qtdeclarative.rs
```diff
@@ -201,6 +202,106 @@ impl Default for QQuickView {
...
+    /// Create a QmlComponent using the QmlEngine.
+    pub fn new(engine: &QmlEngine) -> QmlComponent {
+        unsafe {
+            cpp!([engine as "QmlEngineHolder*"] -> QmlComponent as "QQmlComponentHolder" {
+                return QQmlComponentHolder(engine->engine.get());
+            })
+        }
+    }
+
+    pub fn get_cpp_object(&self) -> *mut c_void {
```
ogoffart: Could also have some documentation.
(Also explain what's the relation with the method in the QObject trait)

qmetaobject/src/qtdeclarative.rs
```diff
@@ -201,6 +202,106 @@ impl Default for QQuickView {
...
+
+    /// Performs QQmlComponent::create
+    pub fn create(&mut self) -> *mut c_void {
+        unsafe { cpp!([self as "QQmlComponentHolder*"] -> *mut c_void as "QObject*" {
+            return self->component->create();
+        })}
+    }
+
+    /// See Qt documentation for QObject::destroyed
+    pub fn status_changed_signal() -> CppSignal<fn(status: ComponentStatus)> {
```
ogoffart: Documentation needs fix.  
I also would like to see this used in a test, please.

agentsim: I've copied ```test_future``` as a template for implementing this test. In so doing I've noticed that both my test and ```test_future``` crash on macOS Catalina with Qt 5.13.1 and nightly Rust.

ogoffart:
There are two problems with the tests:

1. The tests `create_component` and `component_status_changed` are run in parallel in different thread and both create a QmlEngine, which itself create a QApplication.  But there can only be one QApplication per object.  In fact there should only be one instance of QmlEngine.  I realize that the API is unsound and QQmlEngine should better panic! if it tries to instentiate a second one.  But the solution for your test is to move them in the `tests` directory, there you can acquire the `lock_for_test()`

2. You must somehow call `quit` in the QmlEngine otherwise it is gonna run forever.

ogoffart:
Thanks for the contributions!

PR #59, opened 2019-10-31
Avoid using impl dyn

agentsim:
`impl dyn` methods taking `self` by reference imply the `+'static` lifetime on the `self` reference. If the object implementing the trait has a custom lifetime the `impl dyn` methods cannot be used.

It is also illegal to have a generic method on a trait object, so `update_model_indexes` was modified to take the `FnMut` parameter by reference and without using a generic parameter.

### DIFF
--- a/qmetaobject/Cargo.toml
+++ b/qmetaobject/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "qmetaobject"
-version = "0.1.3"
+version = "0.2.0"
 edition = "2018"
 authors = ["Olivier Goffart <ogoffart@woboq.com>"]
 build = "build.rs"
--- a/qmetaobject/Cargo.toml
+++ b/qmetaobject/Cargo.toml
@@ -14,7 +14,7 @@ repository = "https://github.com/woboq/qmetaobject-rs"
 chrono_qdatetime = ["chrono"]
 
 [dependencies]
-qmetaobject_impl = { path = "../qmetaobject_impl", version = "=0.1.3"}
+qmetaobject_impl = { path = "../qmetaobject_impl", version = "=0.2.0"}
 lazy_static = "1.0"
 cpp = "0.5"
 chrono = { version = "0.4", optional = true }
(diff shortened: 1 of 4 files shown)


ogoffart:
Thanks.  I did not really understand what   `impl QAbstractItemModel` really meant.
Yet, i did not really want to have that as method part of the trait because they should not be specialized for each type. But i guess that's still fine.

Does this change require to bump the version to 0.2 according to semver?

agentsim:
> Thanks. I did not really understand what `impl QAbstractItemModel` really meant.
> Yet, i did not really want to have that as method part of the trait because they should not be specialized for each type. But i guess that's still fine.

I'm not a Rust expert (very much still learning!), so I'm really not sure if there is a nicer way, nor am I entirely certain that this causes more specialization than what was there before.

> Does this change require to bump the version to 0.2 according to semver?

See above re: lack of knowledge :)
My guess is yes, because you are probably right about the change in specialization.

agentsim:
Is there anything I can do to help get this merged?

ogoffart:
I was thinking waiting a bit with this merge request and see if there is more incompatible changes.
For example, now that Pin is stable, maybe it would be worth using it somehow.
I believe the way i am using the &mut currently is unsound in some places.

agentsim:
Any thoughts on merging this yet?

rubdos:
(some notes from experimenting on this, to get it back to semver-compatibility)

> > Does this change require to bump the version to 0.2 according to semver?
> 
> See above re: lack of knowledge :)
> My guess is yes, because you are probably right about the change in specialization.

I'm honestly not sure about that, because the impl block change still covers `&mut dyn QAbstractListModel`, does it not? I think you can quite easily test that, just look at whether code that uses that explicit cast still works?

> It is also illegal to have a generic method on a trait object, so update_model_indexes was modified to take the FnMut parameter by reference and without using a generic parameter.

Something like
```diff
-    fn update_model_indexes(&self, f: &mut dyn FnMut(QModelIndex) -> QModelIndex)
+    fn update_model_indexes<F: FnMut(QModelIndex) -> QModelIndex>(&self, mut f: F)
+    where
+        Self: Sized,
```

is still possible though, because calling this on `&mut dyn QALM` would call this on a reference (which is `Sized` afaik). I think the parameter change there, from `FnMut` to `&FnMut` is the only real (semver) breaking change.

Note that that diff still requires to internally mutate to `&mut dyn FnMut`, because the cpp! macro and rust! pseudo-macro don't seem to like to carry the generic parameter over :-(

ogoffart:
Better late than never, this was merged.

PR #63, opened 2019-11-20
Add SimpleListModel::iter()

timsueberkrueb:
This is convenient for iterating over the values in the model.

### DIFF
--- a/qmetaobject/src/listmodel.rs
+++ b/qmetaobject/src/listmodel.rs
@@ -259,6 +259,9 @@ impl<T: SimpleListItem> SimpleListModel<T> {
         self.values = data;
         (self as &mut dyn QAbstractListModel).end_reset_model();
     }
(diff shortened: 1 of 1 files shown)


qmetaobject/src/listmodel.rs
```diff
@@ -259,6 +259,9 @@ impl<T: SimpleListItem> SimpleListModel<T> {
         self.values = data;
         (self as &mut dyn QAbstractListModel).end_reset_model();
     }
+    pub fn iter(&self) -> std::slice::Iter<'_, T> {
```
ogoffart: should it maybe return `impl Iterator<...>`  ?

timsueberkrueb: +1. Can change + add docs and tests tomorrow if you like :)

PR #64, opened 2019-11-21
Added support for passing command line arguments to QmlEngine

Kurble:
Because I needed to pass command line arguments to the `QApplication` instance I've implemented support for it by adding a new constructor `QmlEngine::new_with_args(..)`.

This way it's possible to pass a platform for the qt platform abstraction system, such as `-platform eglfs`. This is required on embedded systems without window manager.

As far as I know this pull request does not introduce any breaking changes.

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -44,29 +44,14 @@ cpp! {{
         }
     };
 
-    struct Arguments {
-        int argc;
-        char **argv;
-
-        ~Arguments() {
-            for (int i=0; i<argc; ++i) {
-                delete[] argv[i];
-            }
-            delete[] argv;
-        }
-    };
-
     struct QmlEngineHolder : SingleApplicationGuard {
         std::unique_ptr<QApplication> app;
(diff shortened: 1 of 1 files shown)


qmetaobject/src/qtdeclarative.rs
```diff
@@ -44,19 +44,29 @@ cpp! {{
         }
     };
 
+    struct Arguments {
+        int argc;
+        char **argv;
+
+        ~Arguments() {
+            for (int i=0; i<argc; ++i) {
+                delete[] argv[i];
```
ogoffart: i don't think it is safe to use the C++ de-allocation function for something that was allocated from Rust.

Kurble: I guess passing the actual arguments from `std::env::args()` would be a better solution, yes. The QT_QPA_PLATFORM environment variable is one that I wansn't able to find somehow, that actually gave me a better solution to what I was trying to achieve, so I don't really need this change anymore. I could still follow through on it with the `std::env::args()` route though, if it's still wanted?

ogoffart:
Thank you!

PR #70, opened 2020-01-15
Wrap all QuickControls 2 methods which do not return QStringList

JBBgameich:


qmetaobject/src/qtquickcontrols2.rs
```diff
@@ -1,9 +1,83 @@
...
+
+    /// Refer to the Qt documentation for QQuickStyle::setFallbackStyle.
+    pub fn set_fallback_style(style: QString) {
+        cpp!(unsafe [style as "QString"] {
+            QQuickStyle::setFallbackStyle(style);
+        })
+    }
+
+    /// Refer to the Qt documentation for QQuickStyle::setStyle.
+    pub fn set_style_with_qstring(style: QString) {
```
JBBgameich: Not sure what to do with this. Does it make any difference to setting the environment variable?

qmetaobject/src/qtquickcontrols2.rs
```diff
@@ -1,9 +1,83 @@
...
+
+use crate::qttypes::QString;
+
+cpp! {{
+    #include <QtQuickControls2/QQuickStyle>
+}}
+
+cpp_class!(
+/// Wrapper around Qt's QQuickStyle class
+    pub unsafe struct QQuickStyle as "QQuickStyle"
```
ogoffart: Everything is static within QQuickStyle, so there is no need to wrap it with cpp_class.  (cpp_class will add the Default, Drop, Clone impl, but we don't need it as we don't need an instance of QQuickStyle)

JBBgameich:
Thanks for the review. I moved the Qt Quick Controls 2 dependency behind a feature now, this way the getters for path and name can be implemented. Using the environment variables would be fine for me as well though. Please let me know which one you prefer.

ogoffart:
the CI fails because addStylePath was only added in Qt 5.12, and I am still supporting Qt 5.9 and even some older version of Qt.
That's another reason to use the env variable as this does not depends on the Qt version.

ratijas:
Qt version could be feature-gated, for sure. But then again, QtQuick.Controls2 may not belong to this repository that directly. After all, QMetaObject is all about more general stuff, and its a huge miracle by itself that it coincided with internal nature of QML.

I'm not saying it couldn't been done as a separate repository under this umbrella workspace, though :thinking:

PR #71, opened 2020-01-16
Fix execute_async

iovxw:
See commits :smile:


ogoffart:
Thanks!
(tests would have been nice)

PR #75, opened 2020-02-19
qml_register singleton

surban:
Registers a QML singleton object.

This has the same purpose as the C++ qmlRegisterSingletonType function but uses qmlRegisterSingletonInstance, which was introduced in Qt 5.14, to avoid the need for a callback.

I am not sure if ownership and destruction are handled correctly, so please check carefully ;)

### DIFF
--- a/qmetaobject/build.rs
+++ b/qmetaobject/build.rs
@@ -17,6 +17,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 extern crate cpp_build;
 use std::process::Command;
+use semver::Version;
 
 fn qmake_query(var: &str) -> String {
     String::from_utf8(
@@ -33,6 +34,7 @@ fn qmake_query(var: &str) -> String {
 fn main() {
     let qt_include_path = qmake_query("QT_INSTALL_HEADERS");
     let qt_library_path = qmake_query("QT_INSTALL_LIBS");
+    let qt_version = qmake_query("QT_VERSION").parse::<Version>().expect("Parsing Qt version failed");
     let mut config = cpp_build::Config::new();
 
     if cfg!(target_os = "macos") {
@@ -74,4 +76,8 @@ fn main() {
         "cargo:rustc-link-lib{}=Qt{}Qml",
         macos_lib_search, macos_lib_framework
     );
+
+    if qt_version >= Version::new(5, 14, 0) {
+        println!("cargo:rustc-cfg=qt_5_14");
+    }
 }
--- a/qmetaobject/build.rs
+++ b/qmetaobject/build.rs
@@ -45,7 +45,6 @@ fn main() {
     }
 
     if qt_version >= Version::new(5, 14, 0) {
-        config.define("QT_5_14", None);
         println!("cargo:rustc-cfg=qt_5_14");
     }
 
--- a/qmetaobject/build.rs
+++ b/qmetaobject/build.rs
@@ -16,8 +16,8 @@ OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 extern crate cpp_build;
-use std::process::Command;
 use semver::Version;
+use std::process::Command;
 
 fn qmake_query(var: &str) -> String {
     String::from_utf8(
@@ -34,7 +34,9 @@ fn qmake_query(var: &str) -> String {
 fn main() {
     let qt_include_path = qmake_query("QT_INSTALL_HEADERS");
     let qt_library_path = qmake_query("QT_INSTALL_LIBS");
-    let qt_version = qmake_query("QT_VERSION").parse::<Version>().expect("Parsing Qt version failed");
+    let qt_version = qmake_query("QT_VERSION")
+        .parse::<Version>()
+        .expect("Parsing Qt version failed");
     let mut config = cpp_build::Config::new();
 
     if cfg!(target_os = "macos") {
@@ -45,7 +47,7 @@ fn main() {
     if qt_version >= Version::new(5, 14, 0) {
         config.define("QT_5_14", None);
         println!("cargo:rustc-cfg=qt_5_14");
-    }    
+    }
 
     config.include(qt_include_path.trim()).build("src/lib.rs");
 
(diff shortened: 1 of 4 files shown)


surban:
Build fails due to Qt version being older than 5.14.

Shall I introduce a feature flag in Cargo.toml for the Qt version?

ogoffart:
Thanks for the pull request!

> Shall I introduce a feature flag in Cargo.toml for the Qt version?

I don't think there is the need for a feature flag, but there can be a cfg that is set from build.rs with something like

```
 println!("cargo:rustc-cfg=qt_5_14");
```

And the version can be known using the output of qmake

ogoffart:
> I am not sure if ownership and destruction are handled correctly, so please check carefully ;)

This looks indeed suspicious.
I wonder if it would not be better to expose `qmlRegisterSingletonType` instead.
We can avoid the template code by using `QQmlPrivate::qmlregister(QQmlPrivate::SingletonRegistration, ...` directly.

surban:
I've added the version check. Will look into `qmlRegisterSingletonType` tomorrow.

surban:
After reading some of the Qt documentation I think that `qmlRegisterSingletonType` and `qmlRegisterSingletonInstance` have two, slightly different, purposes.

- `qmlRegisterSingletonInstance` takes a `QObject` instance that will be shared between all QML engines. The caller owns the `QObject` and must guarantee that it will outlive all `QQmlEngine` instances. This function is useful when the caller wants to control the state of the `QObject` before exposing it as a singleton to QML.
  
  My current implementation of `qmlRegisterSingletonInstance` leaks the Rust object by calling `Box::into_raw()` and returns a `QPointer` so that the caller can access the singleton object but not drop it.

- `qmlRegisterSingletonType` takes a function that will be called to instantiate a singleton `QObject` for each QML engine. The QML engine will own the `QObject` and presumably free it when it shuts down.

  I got around calling directly into Qt internals by instantiating the template parameter as `QObject`. This seems to work nicely, but maybe there are some issues I am not aware of.

ogoffart:
> qmlRegisterSingletonInstance takes a QObject instance 

Right, and the function you implemented does not take a QObject instance.
I think we should mimic that and that the API should take an existing object.  (maybe that is not default constructible)

qmetaobject/src/qtdeclarative.rs
```diff
@@ -436,6 +436,79 @@ pub fn qml_register_type<T: QObject + Default + Sized>(
...
+    };
+    let callback_fn: extern "C" fn(
+        _qml_engine: *mut c_void,
+        _js_engine: *mut c_void,
+    ) -> *mut c_void = callback_fn::<T>;
+
+    unsafe {
+        cpp!([uri_ptr as "const char*", version_major as "int", version_minor as "int",
+                type_name_ptr as "const char*", callback_fn as "QmlRegisterSingletonTypeCallback"] {
+            qmlRegisterSingletonType<QObject>(uri_ptr, version_major, version_minor, type_name_ptr,
```
ogoffart: I think this registers the wrong QMetaObject (the one from QObject).
There is also a non-template overload that should be called instead.

qmetaobject/src/qtdeclarative.rs
```diff
@@ -436,6 +436,79 @@ pub fn qml_register_type<T: QObject + Default + Sized>(
...
+    version_minor: u32,
+    type_name: &std::ffi::CStr,
+    obj: T,
+) {
+    let uri_ptr = uri.as_ptr();
+    let type_name_ptr = type_name.as_ptr();
+
+    let obj_box = Box::new(RefCell::new(obj));
+    let obj_ptr = unsafe { T::cpp_construct(&obj_box) };
+    Box::into_raw(obj_box);
```
ogoffart: The documentation should specify that the object is going to be "leaked".

qmetaobject/src/qtdeclarative.rs
```diff
@@ -436,6 +436,79 @@ pub fn qml_register_type<T: QObject + Default + Sized>(
...
+
+    unsafe {
+        cpp!([uri_ptr as "char*", version_major as "int", version_minor as "int",
+                type_name_ptr as "char*", obj_ptr as "QObject*"] {
+            #ifdef QT_5_14
```
ogoffart: You do not need this macro: you can do 

```suggestion
            #if QT_VERSION >= QT_VERSION_CHECK(5,14,0)
```

qmetaobject/build.rs
```diff
@@ -33,13 +34,21 @@ fn qmake_query(var: &str) -> String {
...
+        .expect("Parsing Qt version failed");
     let mut config = cpp_build::Config::new();
 
     if cfg!(target_os = "macos") {
         config.flag("-F");
         config.flag(qt_library_path.trim());
     }
 
+    if qt_version >= Version::new(5, 14, 0) {
+        config.define("QT_5_14", None);
```
ogoffart: you do not need this.

surban: Changed.

qmetaobject/src/qtdeclarative.rs
surban: Changed.

qmetaobject/src/qtdeclarative.rs
surban: Done.

qmetaobject/src/qtdeclarative.rs
surban: I was hoping to use the overload taking a C++ std::function so that I could accept a Rust closure instead of being forced to Default-construct the object.

But I guess I will have to change it back.

surban: The non-template overload can only be used to register a callback that returns a `QJSValue`.

`inline int qmlRegisterSingletonType(const char *uri, int versionMajor, int versionMinor, const char *typeName, QJSValue (*callback)(QQmlEngine *, QJSEngine *))`

surban: I am now calling functions from `qqmlprivate.h` directly. Sadly there seems to be no other way around the template type issue.

surban:
I've done the adjustments.

surban:
I've discovered one further issue with `qml_register_singleton_type`.

If the default constructor of the singleton object creates a `queued_callback`, it seems that it does not get called when the event loop starts. Presumably the reason for that is that the C++ object has not yet been constructed at the time of callback creation.

To work around this issue I've introduced the `QSingletonInit::init` function which gets called after the C++ object has been constructed. This seems to work well in my use case, but perhaps you have a better idea how to avoid this issue?

ogoffart:
Thanks!  I squashed the commits.

PR #76, opened 2020-02-22
qreal on 32 bit ARM is 32 bit float

rubdos:
It seems that [for some platforms, a `qreal` should be `f32`](https://doc.qt.io/qt-5/qtglobal.html#qreal-typedef).  ~~I'm not sure whether the condition I add here is correct, but at least I get it to compile for SailfishOS with (among others) this patch.~~

EDIT: the edited patch checks whether `qconfig.h` contains a definition of `QT_COORD_TYPE`, which is the only possible redefinition of `qreal`.

I've had the same issue on SailfishOS-i486, which reminded me of this outstanding patch. It doesn't have anything to do with being on ARM.

---

On a completely different note: I've [molested `build.rs` and deleted some features in a fork](https://gitlab.com/rubdos/qmetaobject-sfos-rs/-/tree/sailfishos) to support Qt 5.6 with a cross-compiled MerSDK.  When SailfishOS finally decides to migrate to some modern Qt distribution, I hope to upstream a build system with your repo here, but that should obviously be a bit cleaner.

### DIFF
--- a/qmetaobject/Cargo.toml
+++ b/qmetaobject/Cargo.toml
@@ -12,7 +12,6 @@ repository = "https://github.com/woboq/qmetaobject-rs"
 
 [features]
 chrono_qdatetime = ["chrono"]
-qreal-is-float = []
 
 [dependencies]
 qmetaobject_impl = { path = "../qmetaobject_impl", version = "=0.1.4"}
--- a/qmetaobject/Cargo.toml
+++ b/qmetaobject/Cargo.toml
@@ -12,6 +12,7 @@ repository = "https://github.com/woboq/qmetaobject-rs"
 
 [features]
 chrono_qdatetime = ["chrono"]
+qreal-is-float = []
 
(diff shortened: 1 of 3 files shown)


ogoffart:
I do think the condition is not this one. It depends on how Qt was configured. 

So the way to do that is probably to have a rust feature for it.

It could be detected automatically in build.rs, I think you'd have to search for QT_COORD_TYPE in {qt_include_path}/QtCore/global/qconfig.h

rubdos:
I'll test this on my custom build system and let you know whether it works. I'm not sure how to test this in CI.

rubdos:
The patch works for me. It seems like your CI runs with `--all-features`, so I used `cargo:rustc-cfg=qreal_is_float` instead, after your example! I had no idea how that worked.

If the final result is okay, I suggest I squash the commits together.

ogoffart:
Thanks!

PR #77, opened 2020-02-24
Make qmake binary configurable

semtexzv:
Make `qmake` binary configurable through env variable. This is needed because some linux distros (I'm looking at you fedora) have it under qmake-qt5 name, since base platform QT is 4 and everything from Qt5 must be marked as such.


ogoffart:
Thanks for the patch.

PR #82, opened 2020-04-22
Feature: qrc! visibility modifier + resource prefix + docs

ratijas:
Add visibility modifier to fn generated by qrc!

Because I felt like I'd rather do resources bookkeeping in separate
module, this way qrc!-generated fn should not be private in module.

Also, somehow rewrote `qrc!` public documentation, flattened internal parsing code, fixed minor typos.

In the middle of playing with procedural macros (which was a new shiny thing for me), I stumbled upon a compiler bug. In fact, even two of them and one more in separate library. See https://github.com/rust-lang/rust/issues/71422 for details.


qmetaobject/src/qrc.rs
```diff
@@ -21,39 +21,84 @@ Q_CORE_EXPORT bool qRegisterResourceData(int, const unsigned char *,
...
-///
-/// The macro creates a function that needs to be run in order to register the
-/// resource. Calling the function more than once has no effect.
 #[macro_export]
-macro_rules! qrc { // This is just a forwarding marco so it is documented
-    ($fn_name:ident, $($tokens:tt)* ) => {
-        qrc_internal!($fn_name, $($tokens)*);
-    }
+macro_rules! qrc {
+    // Due to a bug in Rust compiler, empty visibility modifier passed to proc macro breaks parsing
```
ogoffart, line 89: have you tried jsut forwarding `$($tokens:tt)*`  ?

Also, i think it may be now possible to have the qrc macro implemented as a procedural macro, and still get the documentation in the qmetaobject crate.  is there not a new attribute so that the documentation is kept on re-exports?

qmetaobject/main.qml
```diff
@@ -0,0 +1,3 @@
+import QtQuick 2.0
+Item {}
+// This file exists only for doctests
```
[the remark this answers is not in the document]
ratijas: I too, don't like it. But from documentation perspective, it produces cleaner minimal doctest code, i.e. does not require extra subdirectories in path. Unfortunately, there's no way of changing cwd for `cargo test` command. So, a trade-off has to be made.

Please, confirm if you insist on moving files out of create's root directory.

qmetaobject/src/qrc.rs
ratijas: Forwarding tokens is meh. Macro signature would look less descriptive than it could be.

I have no idea about documentation re-exports, need to research on that. As far, as I understood, you want to get rid of intermediate macro call, rename and re-export qrc! macro from proc crate, but retain docs in the non-proc crate, right?

ratijas: Some info on-topic:
https://users.rust-lang.org/t/solved-is-it-possible-to-reexport-a-proc-macro/16864/

Yup, I managed re-export it from proc crate into main crate via `pub use` with new docstring and without `#[macro_blah_blah_blah]` stuff. However, docstring is not retained on re-export of the re-documented re-export.

Consider removing mod qrc or making it private, then re-export and re-document macro in lib.rs?

qmetaobject/src/qrc.rs
```diff
@@ -21,39 +21,84 @@ Q_CORE_EXPORT bool qRegisterResourceData(int, const unsigned char *,
...
-/// ```ignore
-///    <RCC>
-///        <qresource prefix="/qml">
-///            <file>main.qml</file>
-///            <file alias="foo/Foo.qml">Foo.qml</file>
-///        </qresource>
-///    </RCC>
+/// ```xml
+/// <RCC>
+///     <qresource prefix="/qml">
```
ogoffart, line 81: you changed the prefixn it should be foo

qmetaobject/main.qml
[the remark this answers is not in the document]
ogoffart: I don"r insist. But i still would like this to be clean.

Can you make it like this:

```rust
qrc!(my_resource,
     "foo" {
         "qrc_example/main.qml",
         "qrc_example/Bar.qml" as "baz/Foo.qml",
￼      }
￼);
```

otherwise, just leave the comment as ```ignore,  it is should still be tested by other tests.

We could also imagine to pass a path in the qrc! macro,  and it would be hidden with a # for the doc.
```rust
qrc!(my_resource,
# #[path("../tests/qrc_test")]
     "foo" {
```

But i don't want to insist.

qmetaobject/src/qrc.rs
ogoffart: yes, we can re-document the macro in lib.
However if that's a source break, maybe it can wait a bit.

qmetaobject/src/qrc.rs
ratijas: whops

qmetaobject/src/qrc.rs
ratijas: Not that I think about it, breakage is not required. Just re-export and re-document macro in lib.rs, and leave qrc.rs alone with its single internal unsafe fn. I bet, no-one used macro via `qmetaobject::qrc::qrc!` path.

qmetaobject/main.qml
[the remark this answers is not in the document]
ratijas: I like those fresh ideas out of nowhere.

Talking about macro syntax expansion, how about being consistent with file definitions, but at the cost of not being consistent with Qt's *.qrc format?

```rust
qrc!(my_resource,
     "./qrc_example" as "foo" {
         "main.qml",
         "Bar.qml" as "baz/Foo.qml",
      }
);
```

so here,  "./qrc_example" as "foo"  meas directory "./qrc_example" is used as a base path for stuff under prefix "foo".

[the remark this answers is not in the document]
ogoffart: > so here, "./qrc_example" as "foo" meas directory "./qrc_example" is used as a base path for stuff under prefix "foo".

looks good

ratijas:
So, back to work.

My plan is, step by step:

1. fix docstring RCC example
2. move to the root:
  - get rid of the intermediate declarative macro
  - re-export proc macro from the root
  - move documentation there
3. implement resources' prefix dir syntax
4. clean up:
  - move files required for docstring examples somewhere, using new prefix syntax
  - make sure examples for macro and xml are still rendered as equivalent

Sorry, I'm keeping posting, but I feel like I need it for better time & task management.

qmetaobject/src/qrc.rs
ratijas: Resolved by removing the intermediate declarative macro altogether.

qmetaobject/main.qml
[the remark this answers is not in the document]
ratijas: Implemented in 26a0572.

ratijas:
To do: probably the copyright in the new qml files?

qmetaobject/src/lib.rs
```diff
@@ -881,6 +881,87 @@ pub fn install_message_handler(logger: extern "C" fn(QtMsgType, &QMessageLogCont
...
+/// └── src
+///     └── main.rs
+/// ```
+/// then the following Rust code:
+/// ```
+/// # extern crate qmetaobject;
+/// # use qmetaobject::qrc;
+/// qrc!(my_resource,
+/// # // For maintainers: this is actually tested against read files.
+///     "foo" as "tests/qml" {
```
ogoffart, line 939: should it not be the opposite order: `"test/qml" as "foo"`  ?
Makes more sense to have what's after the "as" to be the virtual thing in the ressource.

ratijas: Makes sense. My bad. Was sleepy.

So, the grammar should be like this:

Resource: `[ <physical> as ] <virtual> { File* }`
`<physical> [ as <virtual> ]`

The opposite optional parts.

ratijas:
Do not merge just yet. There's something sketchy going on in resource path resolution.

So far, I documented it with this:

```rust
/// Special string class used to represent Qt strings with pre-computed stable hash.
///
/// Here it is used to represent paths (directories and files) in qrc virtual file system.
#[derive(Debug, Eq, PartialEq, PartialOrd, Ord, Clone)]
struct HashedString {
    hash: u32,
    string: String,
}

/// Virtual file system tree, where leafs (files) are pointing to
/// real files on the local file system.
#[derive(Debug)]
enum TreeNode {
    /// Path to a file on a local file system.
    File(String),
    /// Directory tree for a qrc virtual file system.
    /// Maps qrc virtual paths onto other virtual directories and eventually
    /// onto files on the local file system.
    Directory(BTreeMap<HashedString, TreeNode>, u32),
}
```

ratijas:
Now, all done.

ogoffart:
Thanks for your patches!

I merged the change manually, but somehow github did not close the pull request.
Change 26a05722d67c8788c7db3065b99cb09c697687a1

ratijas:
Hmm, that change is from 19 hours ago. Not everything from this branch made it into master.

ogoffart:
My bad.  (i meged your master branch instead)
Thanks for noticing.

ratijas:
Nice. Time for #84 ?

Issue #85, opened 2020-04-30
Logging facade


ratijas:
Not sure is it just me, or the logging functionality does not work from Qt side by default. I had to copy the handler function from qmetaobject tests, and explicitly install with `install_message_handler` function.

Now, given that there is a message type enum `QtMsgType {QtDebugMsg, QtWarningMsg, ...}` it should be easy to add a function which forwards Qt's messages to the rust logging facade's `info!, warn!, ...`.

ratijas:
[`log!`](https://docs.rs/log/) has a notion of "target" — tag / name / context / module / whatever prefix to figure out where this line is coming from.

Fortunately, there is a notion of a "category" in Qt's logging system, implemented as `const char *` string, and wrapped together with severity (level) in a `QLoggingCategory` class.

When exposed through `qInstallMessageHandler`'s `QtMessageHandler` callback signature, category can be retrieved via `QMessageLogContext::category` member field. And there are already wrappers for that in `qmetaobject`! They even take care on null values. The only thing is, according to Qt source code, default value for a category is (surprisingly) a "default" string, not an empty one.

ratijas:
We've had a big [discussion](https://t.me/rustlang_ru/294964) on Telegram Rust group regarding lifetimes, arguments formatting and  problems related to [`format_args!`](https://doc.rust-lang.org/stable/std/macro.format_args.html) macro, and came to the conclusion that `match` with single all-capturing arm is a common trick to allow retention of temporal objects for "a little bit longer". So, my final version looks like this:

```rust
use log::{Level, logger, Record};
use qmetaobject::*;
use qmetaobject::QtMsgType::*;

fn map_level(lvl: &QtMsgType) -> Level {
    match lvl {
        QtDebugMsg => Level::Debug,
        QtWarningMsg => Level::Warn,
        QtCriticalMsg => Level::Error,
        QtFatalMsg => Level::Error,
        QtInfoMsg => Level::Info,
    }
}

extern "C" fn log_capture(msg_type: QtMsgType,
                          context: &QMessageLogContext,
                          message: &QString) {
    let level = map_level(&msg_type);
    let target = match context.category() {
        "" => "default",
        x => x,
    };
    let file = match context.file() {
        "" => None,
        x => Some(x),
    };
    let line = match context.line() {
        // line numbers start from 1, right?
        0 => None,
        x => Some(x as _),
    };
    let mut record = Record::builder();
    record.level(level)
        .target(target)
        .file(file)
[...]

PR #86, opened 2020-04-30
Forward Qt logging to Rust log!() facade

ratijas:
This is an implementation of proposal #85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,9 +59,33 @@ Window {
 
 Requires Qt >= 5.8
 
+## Cargo features
+
+Cargo provides a way to enable (or disable default) optional [features](https://doc.rust-lang.org/cargo/reference/features.html).
+
+### `log`
+
+By default, Qt's logging system is not initialized, and messages from e.g. QML's `console.log` don't go anywhere.
+The "log" feature enables integration with [`log`](https://crates.io/crates/log) crate, the Rust logging facade.
+
+The feature is enabled by default. To activate it, execute the following code as early as possible in `main()`:
+
+```rust
+fn main() {
+    qmetaobject::log::init_qt_to_rust();
+    // don't forget to set up env_logger or any other logging backend.
+}
+```
+
+### `chrono_qdatetime`
+
+Enables interoperability of `QDate` and `QTime` with Rust [`chrono`](https://crates.io/crates/chrono) package.
+
+This feature is disabled by default.
+
 ## What if a binding for the Qt C++ API you want to use is missing?
 
-It is quite likely that you would like to call a particular Qt function wich is not wrapped by
+It is quite likely that you would like to call a particular Qt function which is not wrapped by
 this crate.
 
 In this case, it is always possible to access C++ directly from your rust code using the cpp! macro.
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ struct Greeter {
 }
 
 fn main() {
+    qmetaobject::log::init_qt_to_rust();
     qml_register_type::<Greeter>(cstr!("Greeter"), 1, 0, cstr!("Greeter"));
     let mut engine = QmlEngine::new();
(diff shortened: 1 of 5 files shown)


ratijas:
I'd like to implement coercion of Qt/Rust logging levels using From/Into traits before merge :thinking:

qmetaobject/Cargo.toml
```diff
@@ -16,6 +16,7 @@ chrono_qdatetime = ["chrono"]
 [dependencies]
 qmetaobject_impl = { path = "../qmetaobject_impl", version = "=0.1.4"}
 lazy_static = "1.0"
+log = "0.4"
```
ogoffart, line 19: can it be made an optional dependency?

qmetaobject/src/log.rs
```diff
@@ -0,0 +1,218 @@
...
+/// Mapping from Qt logging levels to Rust logging facade's levels.
+///
+/// Due to the limited range of levels from both sides,
+/// [`QtCriticalMsg`][`QtMsgType`] and [`QtFatalMsg`][`QtMsgType`]
+/// both map to [`log::Level::Error`][Level],
+/// while [`log::Level::Trace`][Level] is never returned.
+///
+/// [Level]: https://docs.rs/log/0.4.10/log/enum.Level.html
+/// [`QtMsgType`]: https://doc.qt.io/qt-5/qtglobal.html#QtMsgType-enum
+pub fn map_level(lvl: QtMsgType) -> Level {
```
ogoffart, line 85: I wouldn't put these two function `pub`
The From  impls should be enough.

qmetaobject/Cargo.toml
ogoffart: Then it should be documented in the crate's top level documentation in the list of features

ratijas: See https://github.com/woboq/qmetaobject-rs/issues/85#issuecomment-622176232 for my rationale.

But then again, yes, of course I could make it optional.

qmetaobject/src/log.rs
ratijas: OK, reverted back to `From`.

qmetaobject/Cargo.toml
```diff
@@ -11,13 +11,15 @@ keywords = ["Qt", "QML", "QMetaObject",]
 repository = "https://github.com/woboq/qmetaobject-rs"
 
 [features]
+default = ["log"]
```
ogoffart, line 14: Thanks.  I'm just concerned about having too many dependencies.  That's also why i wouldn't put it as the default either.

README.md
```diff
@@ -58,9 +59,33 @@ Window {
 
 Requires Qt >= 5.8
 
+## Cargo features
+
+Cargo provides a way to enable (or disable default) optional [features](https://doc.rust-lang.org/cargo/reference/features.html).
+
+### `log`
+
+By default, Qt's logging system is not initialized, and messages from e.g. QML's `console.log` don't go anywhere.
```
ogoffart, line 68: Don't they go in the stderr by default?

I mean, they follow the default Qt handler which is also quite configurable with env variable: https://doc.qt.io/qt-5/debug.html#warning-and-debugging-messages  /  https://doc.qt.io/qt-5/qloggingcategory.html#configuring-categories

ratijas: Idk, by default they just not showing up at all on my system.

qmetaobject/Cargo.toml
ratijas: I think debugging is essential, especially on such multi-lingua projects, and should be on by default. By having it on by default AND providing instructions for activation on the front-page example we are lowering the barrier for newcomers.

ratijas:
Breaking change incoming:  3da8fe7

PR #87, opened 2020-05-04
log: minor code deduplication

ratijas:
I figured out what was going on inside match expression, and why lifetimes behave so strange.

So, after all, I just factored out two little identical lines into a function, and it still works.


ogoffart:
thanks

Issue #81, opened 2020-04-22
Contributing guidelines, code style


ratijas:
I feel the need for contributing guidelines and code style. Code style should cover both formatting rules (and tools used for automation) and general patterns for hand-written bindings.

For example, I like how it is done in in winapi crate: https://github.com/retep998/winapi-rs/blob/0.3/CONTRIBUTING.md

Things to consider, based on the articles about previous attempts to generate C++ bindings for rust:

 - Method overload renaming rules

    For example, in Qt documentation they are differentiated by suffix number rather than by signature.

     * Constructor overloads:

         + single-argument constructors are expressed using `From` trait.
         + multiple-argument constructors wrapped in dedicated factory function with some sensible name like `impl AB { fn from_a_b(a: A, b: B) -> Self {...} }` — what is the naming pattern for them?
    
     * Method overloads:
        
         + In Rust, method overloading trait-based.
            It works best on single-argument overloads, while multiple arguments may be stacked in tuple.
            For example, see [ToSocketAddrs](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html).

 - Operators overload
    Probably, the easiest part, since in Rust they are generic and single-argumented.

 - Default arguments
    In C++ omitted default arguments are substituted at compile time with what was defined in header file. It can be mapped to `Option`s in Rust. Then, for each `None` argument, a separate `cpp!` macro invocation would need to be done to retrieve the corresponding default value. Finally, call C++ function with all arguments supplied.

 - C-style `enum`s with multiple variants mapping to same values. Sounds like the right job for a `bitflags`-style macro.
    Or just don't use enum at all — that's what winapi is doing: every define becomes just a const.
    Then again, it may be a place for two-layered system, with low-level raw bindings and high-level rusty ones.

 - Short-circuiting on Rust side
    This one is about re-implementing basic things like QRect in native Rust rather than calling into Qt.

     * Is it acceptable?
     * If so, what are the rules?

 - Imitating inheritance?

What else?

ogoffart:
My goal is to make a nice and "idomatic" rust API

> in Qt documentation they are differentiated by suffix number rather than by signature.

I would not consider number as an atractive way to dinstinguish overload.  I think either a name base on the type. Or perhaps taking a trait that can convert from the different expected parameters.

> Default arguments

Either an option,  or no default, or a different overload.

> C-style enums with multiple variants mapping to same values

[...]

PR #89, opened 2020-05-06
qt_doc: Groundwork and showcase

ratijas:
OMG, I'm yet to learn to rebase stuff, sorry for "merge xxx" commits.


ratijas:
OK, that's better.

qmetaobject_impl/src/lib.rs
```diff
@@ -1,4 +1,5 @@
-/* Copyright (C) 2018 Olivier Goffart <ogoffart@woboq.com>
+/* Original work Copyright (C) 2018 Olivier Goffart <ogoffart@woboq.com>
+   Modified work Copyright (C) 2020 ivan tkachenko a.k.a. ratijas <me@ratijas.tk>
```
ogoffart, line 2: Put two "Copyright"   lines,  no need to include Original or Modified

qmetaobject_impl/src/lib.rs
```diff
@@ -26,11 +27,13 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 extern crate syn;
 #[macro_use]
 extern crate quote;
-
+extern crate darling;
```
ogoffart, line 30: I think we no longer need extern crate.

qmetaobject_impl/src/doc.rs
```diff
@@ -0,0 +1,100 @@
...
+
+use darling::FromMeta;
+use proc_macro::TokenStream;
+use syn::{AttributeArgs};
+
+#[derive(Default, FromMeta)]
+#[darling(default)]
+pub struct QtDocMetaArgs {
+    qt: String,
+    // exactly one of them must be present at a time.
```
ogoffart, line 27: I suppose this cannot be an enum because darling won't support it?

qmetaobject_impl/src/doc.rs
```diff
@@ -0,0 +1,100 @@
...
+            + count(&self.cls)
+            + count(&self.method)
+            + count(&self.member)
+            + count(&self.related)
+            + count(&self.typedef)
+    }
+
+    fn kind_and_item(&self) -> (&str, &str) {
+        if self.count_optionals() != 1 {
+            panic!(r#"Exactly one of attributes must be specified: cls, method, member, ..."#);
```
ogoffart, line 51: It would be better to return an Error  so that syn can generate a compile_error! out of it.

(I know that this crate panics because this wasn't supported before)

qmetaobject_impl/src/doc.rs
ratijas: Yeah(((
It would require something like `#[serde(untagged)]` in conjunction with `#[serde(flatten)]` in `darling` world.
It this this structure should be further parsed into a struct with enum, to enforce restriction on a type-level. But not sure how to report errors yet.

qmetaobject_impl/src/lib.rs
ratijas: True. Just tried to be consistent there.

qmetaobject_impl/src/lib.rs
ratijas: About that...
I'm not sure how to do it properly, because I lack any experience in that field.
Copyright sure serves as a pointer, and in a case of emergency it says "that guy may be responsible for it, go ask him/her", but we already have bit blame for that. Also I don't want to get in a way in case of relicensing etc.
Microsoft and some others handle it by requiring contributors to sign [CLA](https://cla.opensource.microsoft.com/MicrosoftDocs/win32) which grants "irrevocable license in the Submission" etc etc etc.

qmetaobject_impl/src/doc.rs
ratijas: OK. I will address error reporting in later PRs but before lending this branch onto master.

After all this macro is all about what _can be done_ with it, rather that what have to be denied. So, graceful recovering from misuses has lower priority, I think.

In the long term outlook, this should become public API, and allow 3rd-party developers wrap parts of Qt modules following upcoming guidelines per #81.

PR #95, opened 2020-05-27
Refactor (cont)

ratijas:
Sorry. There was actually few more little changes which do not deserve much attention, so I'd like to split them from the juicy refactoring hunks.
As always, commit messages are self-explanatory.


ratijas:
What a shame, I missed the rest of qtdeclarative.rs file.

PR #99, opened 2020-06-02
reformat README slightly

ratijas:
copy example from lib.rs docs (should've been done earlier),
fix more whitespaces in C++ types,
added more links.


ratijas:
Such a miserable change, not even worth paying attention to.

Issue #74, opened 2020-02-06
Missing QVariant -> Primitive Type Conversions


Ace4896:
Recently I found that the QVariant wrapper can be created from Rust's primitive types (i32, f32 etc.), but there isn't a way to directly convert it back into a primitive type.

I'd take this on if I have free time, but what I've thought of so far is implementing `TryInto<...>` for the QVariant wrapper (since the conversion can fail).

ogoffart:
You can use the QMetaType trait for that:  there is QMetaType::from_qvariant.

But yes, i agree that having TryInto would also make sense!

Ace4896:
Ah, looks like I completely missed it :sweat_smile: But yeah, I think having `TryInto` would make the type conversion things easier to find.

PR #101, opened 2020-06-05
Add `QVariant::qu?longlong` constructors

rubdos:
Add two constructors for QVariant for int types of 64 bits.


ratijas:
Relates to #54, #74 and ultimately to #81.

ogoffart:
Looks good! Thanks!

PR #103, opened 2020-06-14
Refactor & comments on qobject_impl internals

ratijas:
Refactored some nested code, extracted magic offsets in arguments array
into documented variables. Localized usage of unsafe code in some
places, documented every such usage explaining why it should be
considered safe.

Also just a little bit of cosmetic whitespacing and renaming.

----

I believe, now I better understand the process and the result of MOC-like code generation.

### DIFF
--- a/qmetaobject_impl/src/qobject_impl.rs
+++ b/qmetaobject_impl/src/qobject_impl.rs
@@ -81,6 +81,20 @@ fn builtin_type(ty: &syn::Type) -> u32 {
     }
 }
 
+trait IsVoid {
+    fn is_void(&self) -> bool;
+}
+
+impl IsVoid for syn::Type {
+    fn is_void(&self) -> bool {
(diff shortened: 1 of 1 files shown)


ratijas:
Prioritized for myself, because changes are good, but further and further diverging code base makes it harder to get back on tracks.

ratijas:
It took more effort that it was worth it just to rebase, but no complaining :shrug: 

My ultimate goal about this module is to split parsing and code generation completely, and bind them together through a giant data structure holding all the state. Would it be possible with Tokens and token streams? Would it make it easier to test code gen unit by unit? We'll see.

ogoffart (approved):

ratijas:
Yay! Pull Request which literally celebrated its anniversary few days ago. And I finally got my `lazy!{ as }` around to finish it. That's a big win.

Now I wonder, what about those abandoned Gerrit counterparts from last year…

PR #104, opened 2020-06-14
Remove num operations

ratijas:
If corresponding Gerrit change is accepted, I'll be happy to merge this as well.

https://codereview.qt-project.org/c/qt/qtbase/+/301599

Based on #97, to be merged after it.


ratijas:
This does not make sense anymore nor it ever did. Gerrit change was never accepted, and they are still waiting on me. But even if it gets merged, it would take an eternity or two until it propagates to stable public and adopted release, and then we would have to write even more `#if VERSION() >=` code to ensure all cases are handled without silently going to the default branch. So, as good as my original intentions were, they don't hold water. Closing as invalid.

PR #109, opened 2020-09-06
Switch all packages to Rust 2018 edition

### LINKED ISSUE #108: examples\graph\src\nodes.rs fails to build
On windows with recent rust(1.46) it fails to build

ratijas:
Some packages (2 out of 5, both are examples) were still using default
edition, which is Rust 2015. It started causing problems as soon as
`cpp` crate bumped patch version to 0.5.5 only to start using ::core
paths instead of ::std, which -- without explicitly adding the line
`extern crate core;` to the top of main.rs file -- naturally fails
in Rust 2015.

Since most of the project already using 2018 edition, it was more
reasonable to upgrade edition for these packages, than try to
accommodate `core` in code and documentation.

Fixes #108

### DIFF
--- a/examples/graph/Cargo.toml
+++ b/examples/graph/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "graph"
 version = "0.1.0"
+edition = "2018"
 authors = ["Olivier Goffart <ogoffart@woboq.com>"]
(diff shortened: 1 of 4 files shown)


ogoffart:
thanks

PR #115, opened 2020-10-11
Add QAbstractTableModel support

Korvox:
Wanted to use a TableView in an app so I added QAbstractTableModel.

Kept it parallel with ListModel since there should probably eventually be a merge of code between the abstract models.


ogoffart:
Thanks!

PR #116, opened 2020-10-11
Fix deprication warning for QDateTime from a QDate

Korvox:
Just cleaning up warnings. 

Qt 5.15 deprecated `QDateTime(const QDate &);` and recommends migrating to `QDate::startOfDay()`. Its scheduled for removal in Qt 6.

### DIFF
--- a/qmetaobject/src/qttypes.rs
+++ b/qmetaobject/src/qttypes.rs
@@ -315,7 +315,11 @@ impl QDateTime {
     /// [ctor]: https://doc.qt.io/qt-5/qdatetime.html#QDateTime-1
     pub fn from_date(date: QDate) -> Self {
         cpp!(unsafe [date as "QDate"] -> QDateTime as "QDateTime" {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+            return date.startOfDay();
+#else
             return QDateTime(date);
+#endif
         })
     }
 


ogoffart:
Thanks.

That function was introduced in Qt 5.14,  so it needs some 
`#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)`

ratijas:
Shouldn't we instead of modifying an existing function, add a new one, and put it behind a feature-gate & `#if` directive?

ogoffart:
> Shouldn't we instead of modifying an existing function, add a new one, and put it behind a feature-gate & #if directive?

No, the two function do the same thing, so there should just be one.  Qt deprecated the old way because the new name is a more explicit (therefore better?) name.
But here we also have our own name.   I think `QDateTime::from_date` is fine.  Although maybe `QDateTime::from_date_at_start_of_day` would also work...  or `QDate::start_of_day`  could be also added.

ratijas:
New name in Qt is undeniably better, and it's good to see they get rid of (some of) C++ implicitness.

But we certainly don't want our naming to diverge that much from Qt. `QDateTime::from_date` was an unfortunate required bikeshed, because (a) Rust doesn't support constructor/function overloads, and (b) there's no "constructors" in Rust to begin with, so some descriptive name was chosen. But now that they have normal named method in Qt, is see no reasons to diverge any longer.

But of course compatibility matters, and so something should be done about this deprecated constructor. Like I said — feature-gate on Qt version is an option.

ogoffart:
Thanks.

@ratijas : it is true that one could add a `QDate::start_of_day` function to match Qt's one, and it wouldn't even need feature gate since it can be implemented using the QDateTime constructor for older version of Qt.

ratijas:
Huh, indeed. Somehow, I didn't even think of that.

I hope I'll get back to this later next week ((after I finish fixing aura's documentation and all that groff stuff)). There's a lot things to improve, and first and foremost we need to settle on `#[qt_doc]` attribute macro.

PR #117, opened 2020-10-19
Add WebEngineView support

IShashkin:
Add WebEngineView support and also add a simple example of usage.

Qt is quite modular. This pull request adds new dependency to the web engine module. Maybe it's worth to make its optional via #[cfg(feature = "...")] stuff. What do you think about it?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,11 +5,11 @@ environment:
   - TARGET: x86_64-pc-windows-msvc
     ARCH: amd64
     VS: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat
-    QTDIR: C:\Qt\5.10\msvc2017_64
+    QTDIR: C:\Qt\5.12\msvc2017_64
   - TARGET: i686-pc-windows-msvc
     ARCH: x86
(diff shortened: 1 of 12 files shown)


ogoffart:
Thanks for your contribution.
Yes, we should make it an optional feature, because QWebEngine is quite big.

ratijas:
I wouldn't name that function `webengine_initialize`. Instead, let's use ::namespaces and keep it as close to Qt API as possible.

ogoffart:
Right, let's call the function just `initialize` and let's remove `pub use webengine::*;`  so user would call `webengine::initialize()`

IShashkin:
Thanks for feedback.

> Yes, we should make it an optional feature, because QWebEngine is quite big.

Done

> Right, let's call the function just `initialize` and let's remove `pub use webengine::*;` so user would call `webengine::initialize()`

Done

Look like there is [an issue](https://github.com/appveyor/ci/issues/2607) with QtWebEngine and Qt 5.11 32bit. So I switch Qt version to 5.12 in the appveyor config.

IShashkin:
@ogoffart Should I change/fix anything else?

ogoffart:
Looks good, thanks

PR #118, opened 2020-10-28
fix LLVM ERROR:

lseugene:
Global variable 'qt_pluginMetaData' has an invalid section specifier '.qtmetadata': mach-o section specifier requires a segment and section separated by a comma.

### DIFF
--- a/qmetaobject_impl/src/qobject_impl.rs
+++ b/qmetaobject_impl/src/qobject_impl.rs
@@ -944,10 +944,8 @@ pub fn generate(input: TokenStream, is_qobject: bool) -> TokenStream {
         let plugin_data = qbjs::serialize(&object_data);
         let plugin_data_size = plugin_data.len();
         body = quote! { #body
-            
             #[cfg_attr(target_os = "macos", link_section = "__TEXT,qtmetadata")]
-            #[cfg_attr(not(target_os = "macos"), link_section = ".qtmetadata, section")]
-
+            #[cfg_attr(not(target_os = "macos"), link_section = ".qtmetadata")]
             #[no_mangle]
             #[allow(non_upper_case_globals)]
             pub static qt_pluginMetaData: [u8 ; 20 + #plugin_data_size] = [
--- a/qmetaobject_impl/src/qobject_impl.rs
+++ b/qmetaobject_impl/src/qobject_impl.rs
@@ -944,7 +944,10 @@ pub fn generate(input: TokenStream, is_qobject: bool) -> TokenStream {
         let plugin_data = qbjs::serialize(&object_data);
         let plugin_data_size = plugin_data.len();
         body = quote! { #body
-            #[link_section = ".qtmetadata"]
+            
+            #[cfg_attr(target_os = "macos", link_section = "__TEXT,qtmetadata")]
+            #[cfg_attr(not(target_os = "macos"), link_section = ".qtmetadata, section")]
+
             #[no_mangle]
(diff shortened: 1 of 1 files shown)


ogoffart:
Thanks,
I just tried this on linux, but this does not work. the `examples/qmlextensionplugins/` example no longer works with this commit.

```
file:///[...]/qtdeclarative/examples/qml/qmlextensionplugins/plugins.qml:51 plugin cannot be loaded for module "TimeExample": Failed to extract plugin meta data from '[...]/qmetaobject-rs/target/debug/libqmlqtimeexampleplugin.so'
```

qmetaobject_impl/src/qobject_impl.rs
```diff
@@ -944,7 +944,10 @@ pub fn generate(input: TokenStream, is_qobject: bool) -> TokenStream {
         let plugin_data = qbjs::serialize(&object_data);
         let plugin_data_size = plugin_data.len();
         body = quote! { #body
-            #[link_section = ".qtmetadata"]
+            
+            #[cfg_attr(target_os = "macos", link_section = "__TEXT,qtmetadata")]
+            #[cfg_attr(not(target_os = "macos"), link_section = ".qtmetadata, section")]
+
```
ogoffart, line 950: (small style issue: please remove the extra line breaks)

ratijas:
Hi @lseugene ,

I assume your comment is an error message straight from your linker. I'm not much into linker / scripts, so I have little idea what this commit actually does. Can you, please, provide links and descriptions that would justify your changes? Thanks!

ogoffart:
This is a message from running `qmlscene`  (see the README in the examples/qmlextensionplugins folder)
The Qt plugin loader runtime expects to find the metadata in the `.qtmetadata` section   
(see the loading code in [findPatternUnloaded in qlibrary.cpp](https://code.woboq.org/qt5/qtbase/src/corelib/plugin/qlibrary.cpp.html#_ZL19findPatternUnloadedRK7QStringP15QLibraryPrivate) )

ogoffart:
That said, i only tried linux,  and i also don't know what this fixes. maybe the fix is for mac only?

lseugene:
Hi guys,
Yes, this is a fix for macos target.
But the `not (target_os = "macos")` condition was expected to be backward compatible (linux & windows).
I have tested on macos / windows.
There is no way to check on linux right now.

qmetaobject_impl/src/qobject_impl.rs
```diff
@@ -944,7 +944,10 @@ pub fn generate(input: TokenStream, is_qobject: bool) -> TokenStream {
         let plugin_data = qbjs::serialize(&object_data);
         let plugin_data_size = plugin_data.len();
         body = quote! { #body
-            #[link_section = ".qtmetadata"]
+            
+            #[cfg_attr(target_os = "macos", link_section = "__TEXT,qtmetadata")]
+            #[cfg_attr(not(target_os = "macos"), link_section = ".qtmetadata, section")]
```
ogoffart, line 949:
```suggestion
            #[cfg_attr(not(target_os = "macos"), link_section = ".qtmetadata")]
```

ratijas:
Note to myself: look into GitHub CI, add something other than AppVeyor for non-Windows platforms.

ratijas:
That must be it. Sections, segments at al.

https://developer.apple.com/library/archive/documentation/Performance/Conceptual/CodeFootprint/Articles/MachOOverview.html

PR #120, opened 2020-10-29
Use stable C ABI for plugins

lseugene:


ratijas:
Can you still run `cargo test` on a workspace level after this? I'm not sure, but I remember I encountered some issues with cdylib and testing.

lseugene:
> Can you still run `cargo test` on a workspace level after this? I'm not sure, but I remember I encountered some issues with cdylib and testing.

All tests were passed successfully. (macOS, Windows)
The only thing to keep in mind is that on macOS `QmlEngine::exec()` must run on the main thread.
But `cargo test` starts new threads under the hood.
The current workaround for mac is to run cargo test with --test-threads=1

PR #121, opened 2020-10-29
add QmlEngine::load_url method

lseugene:


ogoffart:
Thanks.
Ideally there should be a test.

PR #123, opened 2020-11-13
Support older Qt versions, down to 5.6

rubdos:
Hi!

I know 5.6 is kinda old (an LTS that has known life from 2016-03-16 until 2019-03-16), but this is still what SailfishOS uses, and will probably be that for a while, until they switch to 5.12 (or hopefully something even more recent).

With the eye on keeping the difference between [my qmetaobject-sfos-rs fork](https://gitlab.com/rubdos/qmetaobject-sfos-rs) and your upstream as small as possible, I'd like to introduce `cfg` flags on all functionality between 5.6 and 5.most-recent. This should not impact upstream, but should provide me with a way more maintainable set of patches.

I hope this is alright for you folks!  If it is, I will start adding the relevant `cfg` flags to all `struct`s and methods that require them.

TODO:

- [ ] Perhaps update the readme for minimum needed Qt version?
- [ ] `git rebase --autosquash`
- [ ] Should we annotate the docs of relevant public functions with their respective Qt versions?
- [ ] Rebase my fork and test whether it works!

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -54,6 +54,20 @@ cpp! {{
             , engine(new QQmlApplicationEngine())
         {}
     };
+
+// Equivalent with QMetaObject::inherits(), for 5.6 and lower.
+    bool qmeta_inherits(const QMetaObject *child, const QMetaObject *check) {
+#if QT_VERSION <= QT_VERSION_CHECK(5, 7, 0)
+        do {
+            if (child == check) {
+                return true;
+            }
+        } while ((child = child->superClass()));
+        return false;
+#else
+        return child->inherits(check);
+#endif
+    }
 }}
 
 cpp_class!(
@@ -428,7 +442,7 @@ pub fn qml_register_type<T: QObject + Default + Sized>(
             meta_object
         );
 
-        int parserStatusCast = meta_object && meta_object->inherits(&QQuickItem::staticMetaObject)
+        int parserStatusCast = meta_object && qmeta_inherits(meta_object, &QQuickItem::staticMetaObject)
             ? QQmlPrivate::StaticCastSelector<QQuickItem, QQmlParserStatus>::cast()
             : -1;
 
@@ -877,7 +891,7 @@ impl QJSValue {
         ] -> *mut c_void as "QObject *" {
             QObject *obj = self->toQObject();
             // FIXME: inheritance?
-            return obj && obj->metaObject()->inherits(mo) ? obj : nullptr;
+            return obj && qmeta_inherits(obj->metaObject(), mo) ? obj : nullptr;
         });
         if obj.is_null() {
             return None;
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -649,6 +649,7 @@ pub fn qml_register_singleton_instance<T: QObject + Sized + Default>(
 /// Refer to the Qt documentation for [qmlRegisterUncreatableMetaObject][qt].
 ///
 /// [qt]: https://doc.qt.io/qt-5/qqmlengine.html#qmlRegisterUncreatableMetaObject
+#[cfg(qt_5_8)]
 pub fn qml_register_enum<T: QEnum>(
     uri: &std::ffi::CStr,
     version_major: u32,
@@ -666,6 +667,7 @@ pub fn qml_register_enum<T: QEnum>(
         version_minor as "int",
         meta_object as "const QMetaObject *"
     ] {
+#if QT_VERSION >= QT_VERSION_CHECK(5, 8, 0)
         qmlRegisterUncreatableMetaObject(
             *meta_object,
             uri_ptr,
@@ -674,6 +676,7 @@ pub fn qml_register_enum<T: QEnum>(
             qml_name_ptr,
             "Access to enums & flags only"
         );
+#endif
     })
 }
 
--- a/qmetaobject/src/qttypes.rs
+++ b/qmetaobject/src/qttypes.rs
@@ -124,9 +124,14 @@ impl QDate {
     pub fn get_y_m_d(&self) -> (i32, i32, i32) {
         let mut res = (0, 0, 0);
         let (ref mut y, ref mut m, ref mut d) = res;
-        cpp!(unsafe [self as "const QDate*", y as "int*", m as "int*", d as "int*"] {
+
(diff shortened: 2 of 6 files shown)


bundyo:
A bit offtopic - where one can read how to use this fork for SFOS development? I've been waiting a long time for Jolla to upgrade Qt, but this might not happen soon and I have an app to rewrite. :)

rubdos:
> A bit offtopic - where one can read how to use this fork for SFOS development? I've been waiting a long time for Jolla to upgrade Qt, but this might not happen soon and I have an app to rewrite. :)

You happen to be on IRC or Matrix? I've been using it for [Whisperfish](https://gitlab.com/rubdos/whisperfish/), which is nearing beta release. You can read how it's done there, but it's currently extremely ugly (and it doesn't work in the official SDK, so I use some... workarounds). Its ugliness is partially resolved by this PR ;-)

qmetaobject/src/qtdeclarative.rs
```diff
@@ -428,7 +428,12 @@ pub fn qml_register_type<T: QObject + Default + Sized>(
             meta_object
         );
 
-        int parserStatusCast = meta_object && meta_object->inherits(&QQuickItem::staticMetaObject)
+#if QT_VERSION >= QT_VERSION_CHECK(5, 7, 0)
+        bool inherits = meta_object->inherits(&QQuickItem::staticMetaObject);
+#else
+        bool inherits = true;
```
ogoffart, line 434: In Qt 5.6 we could make a loop and compere QMetaObject parent.
Probably you can write an helper function that does that.
But well, since this is just for a safety assert, i guess it is ok to skip the test.

bundyo:
Unfortunately IRC and Matrix are not among my channels, I'm using almost everything else. :) I'll consider Matrix.

qmetaobject/src/qtdeclarative.rs
```diff
@@ -877,7 +883,12 @@ impl QJSValue {
         ] -> *mut c_void as "QObject *" {
             QObject *obj = self->toQObject();
             // FIXME: inheritance?
-            return obj && obj->metaObject()->inherits(mo) ? obj : nullptr;
+#if QT_VERSION >= QT_VERSION_CHECK(5, 7, 0)
+        bool inherits = obj->metaObject()->inherits(mo);
+#else
+        bool inherits = true;
```
ogoffart, line 889: Here however, we really need to get the inheritance check, otherwise a variant containing a wrong QObject would cause trouble.

rubdos:
> Unfortunately IRC and Matrix are not among my channels, I'm using almost everything else. :) I'll consider Matrix.

Feel free to send an email, get in contact on the Sailfish forum, or use the fork's issue tracker at https://gitlab.com/rubdos/qmetaobject-sfos-rs/-/issues/, or Twitter, or ... anything really :-)

qmetaobject/src/qttypes.rs
```diff
@@ -124,9 +124,17 @@ impl QDate {
     pub fn get_y_m_d(&self) -> (i32, i32, i32) {
         let mut res = (0, 0, 0);
         let (ref mut y, ref mut m, ref mut d) = res;
+
+        // In version prior to Qt 5.7, this method was marked non-const.
```
ogoffart, line 128: I think it's  fine to keep the 5.6 variant.

qmetaobject/src/qtdeclarative.rs
rubdos: Hmm, this idea just came from my forks patch that just dropped the checks. Hehe. I was lazy at the time.

Do you have any pointers to the Qt docs how I could do that?

qmetaobject/src/qttypes.rs
rubdos: Hmm. While I would be okay with that, I do wonder whether the Rust signature should be `pub fn get_y_m_d(&mut self)` instead for Qt 5.6, if we'd be really pedantic.

qmetaobject/src/qtdeclarative.rs
rubdos: Ah, I [found it](https://doc.qt.io/qt-5/qmetaobject.html#superClass), I think. I'll get on that.

rubdos: Push that in cdc67df. I've tested it by swapping `<=` and `>=` and ran `cargo test` on a Qt 5.15 installation, haven't tested yet on my phone, but I don't expect anything weird.

bundyo: You can probably pass all inherits checks through that function and do only one version check inside it. :)

rubdos: Was thinking that too, and I dismissed it because of a reasoning that I didn't make well. I'll change it!

rubdos: 8590a14e5ec8f088f48274d59cb18b93f91775c7

rubdos:
8590a14, 0523076 are related to how simple `cpp_build` is: it cannot exclude code from compilation merely based on a `#[cfg(qt_5_7)]` flag; it will instead include everything and leave it be to `lto`.

qmetaobject/src/qttypes.rs
rubdos: I've kept the non-const version.

rubdos:
Guarding C++ code with `cfg()` generated by `build.rs` is sadly impossible, so I've done it a bit hacky (5419d84). Suggestions welcome.  I've squashed the other commits, as these were slight fixups, and addressed most other comments.

ogoffart:
looks good to me.  Could you rebase?

rubdos:
> looks good to me. Could you rebase?

done. I still have some other Sailfish related issues, but I'll do them separate.

ogoffart:
thanks!

rubdos:
Seems that my Travis addon points to some other problem, cfr. https://travis-ci.org/github/woboq/qmetaobject-rs/jobs/743931414

```diff
diff --git a/qmetaobject/tests/tests.rs b/qmetaobject/tests/tests.rs
index d1a8d9b..8b0acfa 100644
--- a/qmetaobject/tests/tests.rs
+++ b/qmetaobject/tests/tests.rs
@@ -720,6 +720,7 @@ struct MyEnumObject {
     base: qt_base_class!(trait QObject),
 }

+#[cfg(qt_5_7)]
 #[test]
 fn enum_properties() {
     qml_register_enum::<MyEnum>(
```

should do it.

PR #124, opened 2020-11-19
Set correct debug flag in plugin's metadata

lseugene:


ogoffart:
Thanks for the patch.
I think that's correct.
What platform are you trying that on which has problem?
Would it be possible to write a test?

lseugene:
> What platform are you trying that on which has problem?

This is a runtime check when loading plugins on Windows platform.
https://github.com/qt/qtbase/blob/5.15.1/src/corelib/plugin/qlibrary.cpp#L810
https://github.com/qt/qtbase/blob/5.15.1/src/corelib/plugin/qlibrary.cpp#L80

> Would it be possible to write a test?

I don't know how to implement such runtime test "cheaply" with cargo test infrastructure. 
Maybe it would be better to just implement the missing 'qbjs::Value' deserialize part and plain unit tests for it, but I have not found complete encoding format spec to do this.

ogoffart:
Tahnks!

Issue #90, opened 2020-05-10
Research: include_bytes! in qrc! macro?


ratijas:
This is a tracking issue for a planned research activity.

### Tasks

Investigate the possibility to use `include_bytes!`, `syn::LitByteStr` or other methods to include raw binary data in `qrc!` macro implementation. Compare results for the following criteria:
 - Size on disk
 - Overall compilation time
 - Incremental compilation support

### Why?

Few years ago I had a project with a lot of big multimedia files. It was quite painful to recompile about 2GB of resources, especially because it generated even more (x6 times) trash shored on disk. All those `0xYZ, ` which is 6 chars per one byte of an input resource.

So, ideally, for files which are not compressed, `include_bytes!` should be a best fit, I suppose. The question is, how those includes would look like in a larger byte stream (payload) in qrc's generated code.

ratijas:
Thanks, @ZainlessBrombie!

PR #128, opened 2021-02-06
Drastic speedup of qrc macro / file inclusion

ZainlessBrombie:
Currently, the qrc resources are included as a byte array, generating a token stream manually. This is very slow, taking about 20 seconds in my project. This PR instead uses a bytestring literal token, which drastically improves compilation time to the point of no noticable slowdown. As far as I can tell there are no drawbacks to this method.


ZainlessBrombie:
Note that I also ran rustfmt on the file.

ZainlessBrombie:
Closes #90

ogoffart:
Thanks!

PR #130, opened 2021-04-09
Implement QMetaType for QJSValue

IShashkin:
Now QJSValue can be used in arguments of methods and properties.


ogoffart:
Thanks!

PR #132, opened 2021-04-19
Added add_image_provider method to qml engine

zethra:
There currently isn't a way to construct a QQuickImageProvider in Rust but without this method there is no way to use one at all, even defined in cpp.

I've added this because I need for a project. I'd be nice to add a way to construct a QQuickImageProvider in Rust but that's beyond me.

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -232,6 +232,18 @@ impl QmlEngine {
             self->engine->addImportPath(path);
         })
     }
+
+    /// Add an QQuickImageProvider to the QML engine
+    ///
+    /// There currently isn't a way to construct a QQuickImageProvider in
+    /// rust so it is necessary to create a class that extend QQuickImageProvider
+    /// in cpp, then construct one and pass it to this method.
+    pub fn add_image_provider(&mut self, provider_id: QString, provider: *mut c_void) {
+        cpp!(unsafe [self as "QmlEngineHolder*", provider_id as "QString", provider as "QQmlImageProviderBase*"] {
+            self->engine->addImageProvider(provider_id, provider);
+        })
+    }
+
 }
 
 /// Bindings to a QQuickView


ogoffart (changes requested):
Thanks for your contribution!

Howerver i'm afraid this can't be done like that.

The function takes a raw pointer. As such, it should be marked as `unsafe` because it would cause a crash to pass a dangling pointer.

But this show that we don't really want this as is, because, as you said, it would be better to have a pure rust way to add the  QQuickImageProvider.
That would be a bigger work and i understand that you are not ready to contribute that.

What I suggest instead would be to expose a function such as `fn cpp_ptr(&self) -> *mut c_void` to QQmlEngine that return a pointer to the C++ `QQmlEngine`  so you can call the `addImageProvider` function yourself in the code base which also creates the image provider itself.
I believe this would solve your problem, and would also allow many more extensions for other use cases which are currently not exposed by this crate.

ratijas:
Image providers and icons are such a messed up pain in Qt/QML, I could still feel that from a year ago, haha

I think, I did something similar to what @ogoffart described, but for wrapping a different package: get a raw pointer to the engine, and do whatever with it. This [QuickFlux wrapper](https://github.com/InnopolisUniversity/innometrics-agent-os-linux/commit/4b7221fa8f778ea57320a4527c251db148f0cd78) is just an example idea of what you could do.

ogoffart:
@ratijas   ouch, i see in your QuickFlux wrapper that you copied the QmlEngineHolder into your own C++ file. But that sounds like a horrible hack :-/   I believe providing a (rust) `QmlEngine::cpp_ptr`  that returns a pointer should be indeed more appropriate for your use case as well.

ogoffart:
Commit 5d6b0ff added a function to get the raw pointer to QQmlEngine so you can use it in your application.

I'm closing this issue because this should allow you to do what you want.

If you really want to add a image provider api to this crate, it should be a purely rust one.

But thanks anyway for the contribution.

ratijas:
> @ratijas ouch, i see in your QuickFlux wrapper that you copied the QmlEngineHolder into your own C++ file. But that sounds like a horrible hack :-/ 

Oh yeah, that repo is one huge pile of terrifying hacks. Luckily though, that was just a university project, and never made it into production. But, I mean… just imagine all the fun I have had coping with all that stuff.

I don't advice following my path, like in "don't try that at home". But just to get an idea of what's generally possible, it is good enough.

Pure rust handwritten wrappers are the way to go, provided that we have necessary foundation to build upon.

timsueberkrueb:
Hey there, I created an image provider API for my side project: [source code](https://github.com/timsueberkrueb/webber/blob/master/src/qt/image_provider.rs). Does that look like something that could be upstreamed? Thanks!

PR #135, opened 2021-04-22
Cosmit for future

ratijas:
Hey ratijas, that's a nice little stash of commits you've got here. It would be a shame if anything were to... (proceeds to force-push origin master)

### DIFF
--- a/qmetaobject/src/future.rs
+++ b/qmetaobject/src/future.rs
@@ -5,7 +5,6 @@ use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};
 
 use crate::connections::SignalArgArrayToTuple;
 
-
 static QT_WAKER_VTABLE: RawWakerVTable = RawWakerVTable::new(
     |s: *const ()| {
         RawWaker::new(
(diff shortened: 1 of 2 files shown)


ratijas:
Apparently, AppVeyor is not very reliable.  Either that, or GitHub freaked up. Just what on earth was that "failed to update crates.io index"? Anyways, retrying with new changes that cargo fmt didn't catch (namely, inside rust! wrapped in cpp! macros)...

ratijas:
yeah, this is pretty much just a cosmetic change.  I have some more fundamental stuff on the way,  but it needs some rebasing on latest master. Have been skimming through it today , but then suddenly day-job happened 😅

stay tuned but don't worry about breaking stuff -- i'll adapt mine as needed.

ratijas:
Since it's not much of a change, I guess I could just silently merge it like nothing special. I really like how those extra `use`s reduce cognitive load when reading e.g. `Poll` instead of `std::task::Poll`; and those interned unsafe keywords in `cpp!` look more on topic and in-line with overall style of the code base.

ogoffart:
:+1:

PR #136, opened 2021-04-23
feat(qttypes): Do more wrappers around QVariant

ratijas:
Wrapped converters to and constructors from all available types. The rest
of them are marked as TODO comments for future implementers.
Methods and constructors, which I'm not sure about whether they are needed
to be exposed in Rust at all, are marked with `TODO?:` prefix.

Constructors are reordered according to the official Qt documentation
for ease of code navigation. Return types are shortened to `Self`.

Fixed some misleading links to upstream docs in From<{i,u}{32,64}> impls.

Some wrappers are impossible to write at this point without introducing
QMetaType and its Type enum in this crate or its dependencies, as currently
it would create a circular deps between qmetaobject and qttypes crates.
Such hypothetical implementations are left commented out for future ideas.

---

I acknowledge that this is some controversial big boy commit droppin' here like a bomb. I ask to carefully review it, and discuss key design concerns outlined below.

1. Wouldn't it be better if all the `QVariant::toSomething()` methods return `Option<Something>` in rust, by checking for `canConvert()` first? Would be much more Rusty way, at the cost of deviating from Qt origins.
2. `QVariant(const Something &val)` constructors actually take values by references, they never really consume original value. Wouldn't it make more sense to add `&` to all those `impl From<Something> for QVariant` i.e. `impl From<&Something> for QVariant`?
3. Comments with links are missing from implicitly generated basic traits by `cpp!` macro.  No idea what to do.
4. Do we really want all those TODOs stay there probably forever?


qttypes/src/lib.rs
```diff
@@ -640,23 +641,105 @@ impl std::fmt::Debug for QString {
     }
 }
 
+// XXX: The Default, Clone and Copy traits are provided by cpp_class! but they won't have these
+// nice "Wrapper around ..." docs with links on them. How can we fix this?
```
ogoffart, line 645: I don't think we need such documentation. It is obvious for any Rust developer what Default and Clone do. Also the documentation for trait isn't even well displayed in docs.rs

qttypes/src/lib.rs
```diff
@@ -640,23 +641,105 @@ impl std::fmt::Debug for QString {
...
-            return self->toByteArray();
+    /// [method]: https://doc.qt.io/qt-5/qvariant.html#canConvert
+    pub fn can_convert(&self, target_type_id: i32) -> bool {
+        cpp!(unsafe [self as "const QVariant*", target_type_id as "int"] -> bool as "bool" {
+            return self->canConvert(target_type_id);
+        })
+    }
+
+    // TODO: for this to work, we need QMetaType be available in qttypes package.
+    // TODO: It would create a dependency cycle at this point.
```
ogoffart, line 665: I think we don't need to expose this function in this crate.

The function in the QMetaType trait should be enough.

qttypes/src/lib.rs
```diff
@@ -640,23 +641,105 @@ impl std::fmt::Debug for QString {
...
+    /// Wrapper around [`convert(int targetTypeId)`][method] method.
+    ///
+    /// [method]: https://doc.qt.io/qt-5/qvariant.html#convert
+    pub fn convert(&mut self, target_type_id: i32) -> bool {
+        cpp!(unsafe [self as "QVariant*", target_type_id as "int"] -> bool as "bool" {
+            return self->convert(target_type_id);
+        })
+    }
+
+    // TODO?: [static] template <typename Types> QVariant QVariant::fromStdVariant(const std::variant<Types...> &value)
```
ogoffart, line 691: In general, we don't have to replicate the whole API, only the API that make sense.  This clearly make no sense in rust.
Same for the function that operates on the QMetaType which are better done by the QMetaType trait.

qttypes/src/lib.rs
```diff
@@ -640,23 +641,105 @@ impl std::fmt::Debug for QString {
...
+            return self->isValid();
         })
     }
 
+    // XXX: for setValue() see fromValue() comment above.
+
+    /// Wrapper around [`swap(QVariant &other)`][method] method.
+    ///
+    /// [method]: https://doc.qt.io/qt-5/qvariant.html#swap
+    pub fn swap(&mut self, other: &mut QVariant) {
```
ogoffart, line 731: This function also doesn't make sense in rust. One would use  the std::mem::swap function.

qttypes/src/lib.rs
```diff
@@ -666,6 +749,265 @@ impl QVariant {
...
+
+    /// Returns the variant as a Rust's native [`Vec`]`<`[`u8`]`>` if the variant is supported by
+    /// [`to_qbytearray()`] method, otherwise returns an empty vector.
+    ///
+    /// Unlike [`to_qbytearray()`] which might use [implicit sharing] (copy-on-write), this method
+    /// always allocates memory for a [`Vec`].
+    ///
+    /// [`to_qbytearray()`]: Self::to_qbytearray()
+    /// [implicit sharing]: https://doc.qt.io/qt-5/implicit-sharing.html
+    pub fn to_bytes(&self) -> Vec<u8> {
```
ogoffart, line 769: Is there a particular use case for this function?

qttypes/src/lib.rs
```diff
@@ -666,6 +749,265 @@ impl QVariant {
...
+    pub fn to_date_time(&self) -> QDateTime {
+        cpp!(unsafe [self as "const QVariant*"] -> QDateTime as "QDateTime" {
+            return self->toDateTime();
+        })
+    }
+
+    /// Wrapper around [`toDouble(bool *ok)`][method] method.
+    ///
+    /// [method]: https://doc.qt.io/qt-5/qvariant.html#toDouble
+    pub fn to_double(&self) -> Option<f64> {
```
ogoffart, line 797: I guess i'd call this `to_f64` instread

ogoffart (commented):
I only made a partial review so far.

In general, i don't think it is worth it to try to exhaustively wrap every call and expose the full API.
The philosophy behind this whole project is to come up with a idiomatic API to use QML from Rust without having to write C++ or learn Qt. And to do so in a most efficient way. 
Yes, we also want to keep close to the Qt API and i understand this is a conflicting goal, but in doubt, better have idiomatic rust API than blindly imitate Qt.
For example, all these conversion might not be worth it when there is the QMetaType trait doing them. And also the From and Into trait.

> Wouldn't it be better if all the QVariant::toSomething() methods return Option<Something> in rust, by checking for canConvert() first? Would be much more Rusty way, at the cost of deviating from Qt origins.

Yes, i agree, it should

> QVariant(const Something &val) constructors actually take values by references, they never really consume original value. Wouldn't it make more sense to add & to all those impl From<Something> for QVariant i.e. impl From<&Something> for QVariant?

It all depends if the function can sonsume (take ownership) of the memory or not. If not, then it might indeed be better to take value by reference.

> Comments with links are missing from implicitly generated basic traits by cpp! macro. No idea what to do.

I think that's fine. These links wouldn't really help anyone IMHO.

> Do we really want all those TODOs stay there probably forever?

No.

qttypes/src/lib.rs
ratijas: agreed.

qttypes/src/lib.rs
ratijas: yeah, kinda thought about std::mem::swap as well.  It was so weird to work with reference to a pointer passed through FFI, that I even wrote a test to make sure everything matched up.

qttypes/src/lib.rs
ratijas: then, it would apply to all the others in family as well

qttypes/src/lib.rs
ratijas: While is true that redefinition of trait docs do not show up unless you explicitly expand them (+), nevertheless they are still there, safe and sound (no pun intended).


> I think that's fine. These links wouldn't really help anyone IMHO.

Links may not be _that_ important, but (as I mentioned in some other thread) literal copy-pasted signature is essential for someone who looks for a Rust replacement of a particular piece of Qt API. It would be much less of a problem, it things were possible to map 1:1, but instead in Rust we have From traits, absence of constructors and default arguments, different names for basic types (double vs f64) etc. — all of which causes difficulties in navigation.

qttypes/src/lib.rs
ratijas: OK

qttypes/src/lib.rs
ratijas: Idk, just a shortcut, a convenience. Kinda making it more Rust-oriented and usable for non-Qt code.

I wanted to name it `to_vec`, but it would cause all sorts of conflicts with `toList() -> QVariantList`.

ratijas:
Overall, a big thanks for such a fast review. I'm going to call it a day for now, re-think about it overnight, and come up with (suggested) changes tomorrow.

Side note: As much as I'd like to conduct it in Gerrit-style, GitHub doesn't allow viewing inter-patchset diff, so I'll be adding new commits, but manually squash before final merge.

qttypes/src/lib.rs
ratijas: Without assuming much knowledge of Qt from the users of this crate, I'd be happy to mention in Clone docs about implicit sharing of the majority of datatypes directly supported by QVariant. But now that I think about it, such info might fit in QVariant's main doc comment itself.

ratijas:
I've finally had a true weekend, phew. So didn't have a chance to code, but gave it a few thoughts.

I totally agree that in order to make it the Rust way, all those to_something() methods should return an optional value. In fact, they can be implemented as TryInto trait with unit `()` error type, and we should teach folks to use `.try_into().unwrap_or_default()` if they insist on Qt-like behavior.  Such documentation shall be located at Wrapper-specific section of the QVariant type itself.

Once thing that still bothers me, is that `canConvert()` does not guarantee the success of actual conversion, like in String -> UInt directions:

```
let mut v = QVariant::from(QString::from("Yahaha!"));
let uint_type = 3_i32;
// According to QVariant::canConvert(int) rules, QString can be converted to UInt,
assert!(v.can_convert(uint_type));
// but result of such an attempt may fail miserably.
assert_eq!(v.to_uint(), None);
```

which means, for those methods, it might be reasonable to check either both or only `bool *ok`, but not only `canConvert()` result.

Will try my best to push updates ASAP.

ratijas:
Wanted to implement `to_hash_map()` as a feature parity for `to_bytes() -> Vec<u8>`. But it ended up looking so scary, I gave up all hopes on passing any sane code review. Here it is, for archeologists with a lot of historical interest.

```rust
pub fn to_hash_map(&self) -> HashMap<String, QVariant> {
    // QVariant::toHash() is guaranteed to be cheap, fast and constant time O(1).
    // If `self` is not a QHash, an empty one will be returned.
    let size = cpp!(unsafe [self as "const QVariant*"] -> i32 as "int" {
        return self->toHash().size();
    });
    let mut map = HashMap::with_capacity(size as usize);
    let m = &mut map; // Rust's HashMap is opaque for C++
    cpp!(unsafe [self as "const QVariant*", m as "void*"] {
        using namespace std;
        QVariantHash hash = self->toHash();
        QHashIterator<QString, QVariant> i(hash);
        while (i.hasNext()) {
            i.next();
            QString key = i.key();
            QVariant value = i.value();
            rust!(QVariant_toHash_insert [
                m: &mut HashMap<String, QVariant> as "void*",
                key: &QString as "QString&",
                value: &QVariant as "QVariant&"
            ] {
                m.insert(key.clone().into(), value.clone());
            });
        }
    });
    map
}
```

ratijas:
For now, it was only a rebase. Not worth looking into.

flying-sheep:
aw, too bad this isn’t in. Could have used the invalid variant

PR #137, opened 2021-04-27
Shorten paths and modernize imports

ratijas:
Most noticeably, std::ffi::CStr[ing] and ::std::os::raw::c_void.

---

Just a clean-up, because I needed QMetaType in a good shape.

The more I play with QVariant implementations, the more I think QMetaType belongs to qttypes crate. I can not justify that they are not split in different packages.


qmetaobject/src/qtdeclarative.rs
```diff
@@ -975,14 +973,15 @@ mod qjsvalue_tests {
...
+/// # use std::ffi::CStr;
 /// #[derive(Default, QObject)]
 /// struct QExampleQmlPlugin {
 ///     base: qt_base_class!(trait QQmlExtensionPlugin),
 ///     plugin: qt_plugin!("org.qt-project.Qt.QQmlExtensionInterface/1.0"),
 /// }
 ///
 /// impl QQmlExtensionPlugin for QExampleQmlPlugin {
-///     fn register_types(&mut self, uri: &std::ffi::CStr) {
+///     fn register_types(&mut self, uri: &CStr) {
```
ogoffart, line 984: I think it's useful in the example to see what exactly CStr is (that it is std::ffi::CStr and not some other things)

ratijas: Let me see.  I expect docs generator to be smart enough to link it to std::ffi::Cstr.

ratijas: No, unfortunately, the are no clickable words in example blocks whatsoever. WTF, rust? Next things next, let me check what is the official position of std docs regarding this point.

ratijas: Looking at [std::ffi](https://doc.rust-lang.org/std/ffi/index.html) docs, they are explicitly and without hiding behind hashtags `#` *using* their items in each and every example. And they don't bother hiding things imported from other modules, as can be seen in e.g. [std::time::SystemTime](https://doc.rust-lang.org/std/time/struct.SystemTime.html).

Maybe we should try too.

ratijas: The general rule of de-facto seems to be like this:

1. Show your imports explicitly,
2. One crate per line, use braces to group items,
3. Std is an exception to the rule 2: its second level modules may be written one per line, as if they were separate crates.
4. Alphabetic ordering, std has no privileges to be on top of others.
5. No grouping of imports, no blank lines between std/libs/workspace/`crate`/`self` etc. That kind of stuff IntelliJ-Rust absolutely **loves** doing while optimizing imports.

ratijas: Also, `# extern crate qmetaobject;` is not needed at all, since rust doctest inserts it automatically anyway, and `#[macro_use] extern crate qmetaobject;` is no longer needed, since we switched to 2018 edition a while ago.

ratijas:
This is already turning out bigger than anticipated :sweat_smile:

qmetaobject/src/qtdeclarative.rs
ratijas: Some links to back up my claims:

- https://doc.rust-lang.org/rustdoc/documentation-tests.html#pre-processing-examples rules regarding inserting `extern crate`
- https://docs.rs/async-compat/0.2.1/async_compat/struct.Compat.html (sorted uses)
- https://docs.rs/bimap/0.6.1/bimap/hash/struct.BiHashMap.html#method.with_hashers (std on top, but otherwise is fine)

ratijas:
@ogoffart  Do you like it more this way? Looks simple and sparse to me. Very copy-pastable for a reader.


It currently lacks my favorite "wrapper around" and links boilerplate, but that's a different topic.

ogoffart:
TBH i liked it better before (with the full path in the code).  But that's not so important.

examples/graph/src/main.rs
```diff
@@ -1,14 +1,14 @@
 #![allow(non_snake_case)]
 #![allow(unused_variables)]
-#[macro_use]
-extern crate qmetaobject;
-use qmetaobject::scenegraph::*;
-use qmetaobject::*;
+
 #[macro_use]
 extern crate cstr;
```
ogoffart, line 5: yeah, all these extern crate could also be cleaned up, but that's unrelated.

ratijas: I… kinda missed out when cpp and cstr became 2018 edition-compatible. Either that or qmetaobject-rs was at 2015. Anyway, I remember clear as day, it didn't work without `#[macro_use] extern crate`.

Working on a patch...

ratijas:
Ahem… So… Not sure it was worth it in case of `cpp!` in every file, but oh well. Here it goes.

ratijas:
Just a little tiny whitespace nit while I'm at it. Now I consider it done, although I've almost sure missed something. And that's why I'm genuinely uncomfortable with merging it without an approval. There is no rush though. Take your time.

ratijas:
One week was just about enough, I think. Gotta review myself and merge it tomorrow, if no concerns will be raised in the process or by that time.

ratijas:
Shortened more paths, especially in tests. Removed some strange std stuff, such as `std::string::String` and `use std::iter::Iterator;`
`

ratijas:
CI is freaking out.  Retrying.

ratijas:
Done and done. Now, back to QVariant and QMetaType at #136.

PR #139, opened 2021-05-05
Allow overriding Qt header and lib paths with environment variables in qttypes/build.rs

tnibler:
First checking whether the DEP_QT_INCLUDE_PATH and
DEP_QT_LIBRARY_PATH allows compiling against different
Qt versions than whatever qmake picks up on.

This is useful for example when cross-compiling.

### DIFF
--- a/qttypes/build.rs
+++ b/qttypes/build.rs
@@ -72,8 +72,14 @@ fn main() {
             }
         }
     };
-    let qt_include_path = qmake_query("QT_INSTALL_HEADERS").unwrap();
-    let qt_library_path = qmake_query("QT_INSTALL_LIBS").unwrap();
+    let qt_include_path = match std::env::var("DEP_QT_INCLUDE_PATH") {
+        Ok(p) => p,
+        Err(_) => qmake_query("QT_INSTALL_HEADERS").unwrap(),
+    };
+    let qt_library_path = match std::env::var("DEP_QT_LIBRARY_PATH") {
+        Ok(p) => p,
+        Err(_) => qmake_query("QT_INSTALL_LIBS").unwrap(),
+    };
 
     let mut config = cpp_build::Config::new();
 


ogoffart:
I'm trying to understand how do you use that when cross compiling.
Shouldn't the qmake used to cross compile already provide the right settings?
you can override the qmake executable with the QMAKE env variable already.
normally, a Qt build for cross compiling should still a set of host tools on the host platform, and that qmake should provide links to the library for that platform.

tnibler:
This is what I'm doing: https://gitlab.com/cryptocam/cryptocam-companion/-/blob/windows-cxc/WindowsCrosscompile.Dockerfile

I can't manage to set values like `QT_INSTALL_LIBS` in qmake (`qmake -qt=qt5 -set QT_INSTALL_LIBS /path/to/libs` doesn't seem to do anything). If my approach is completely misguided I apologize, I just want some kind of working windows build.

ogoffart:
what i mean is that you should tell the build script to use the cross-compiling qmake with the QMAKE env variable.

For example:

```docker
env QMAKE=/build/qt_libs/5.15.0/mingw81_64/bin/qmake
```
or something like that. 
Then the build script should call the right qmake that has the path to do the cross compilation, i think.
```

tnibler:
The `qmake` from the target platform's Qt is a windows exe though, so I can't use that..

ogoffart:
Ah, that's a good point.
I've crosscompiled when Qt was compiled with the `-xplatform` configure flag. But this is not the case here.

I wonder what is the recommended workflow with qmake to cross-compile like this.

So i guess this change makes total sense then.
I'm just not sure about the name of these environment variable. `DEP_` are environment variable usually set by cargo itself, so it is a bit strange to re-use that (they are "output" rather than "input")
I wonder what's the name of these in cmake for example, or with other build systems.

tnibler:
> `DEP_` are environment variable usually set by cargo itself, so it is a bit strange to re-use that

Oh ok I didn't know that, that name should definitely change then.

As for recommended qmake workflows, I'm of no help there. I avoid it wherever I can. I tried coaxing the host system qmake into pointing to the correct windows qt tree, but it didn't work.

tronical:
I don’t think this way of taking Qt binaries is supported by Qt in general. A build of Qt, with a functioning qmake binary, that runs on Linux but creates win32 binaries can be created by running configure on Linux with -xplatform win32-g++ . There are a few blog posts out there guiding through those steps. Then the detection in this crate should work, I think.

ogoffart:
> I don’t think this way of taking Qt binaries is supported by Qt in general

On the other hand, even if it is not officially supported by Qt, it should still work, and it would be nice to allow it in this crate.
We just need to find a good way to detect the library dir, the include path, and the Qt Version.
Perhaps if qmake can't solve it, we could try to look for some cmake files instead. and then it would just be a matter to pass one environment variable with the Qt path.

ratijas:
Extending on the point of official support.

Qmake is going to be supported for a while, but official Qt company's position is to move everything to CMake. At least, the latest minor is 5.x branch is provided with CMake by default.

Wouldn't it be nice (if possible) to disable qmake interactions whatsoever, and switch to CMake-based solution instead? A feature flag would be required, since we probably can't drop qmake for older Qt 5.x version.

ogoffart:
We use qmake only to query the Qt installation about its path and version.
I just attempted to try to the the same with cmake, but it seems to be a dead end.
Parsing `.cmake` files seems to be way to complicated. I did not get to make `cmake --find-package` to work, also the documentation says it is obsolete. `find_package(Qt5)` in a cmake script run with `cmake -P` also don't work as it errors with `add_library command is not scriptable`  
Maybe one option could be to simulate a build and add a bunch of `message` and parse that output, but that also seem overkill.

So i think maybe the best would simply be the suggestion from this patch, to provide three (and all three) env variables for version/include_path/lib_path.  Maybe one can deduce the version from the include path by parsing the `QtCore/qtcoreversion.h`
We just have to come up with good name for it.

ratijas:
CMake -P script mode is basically useless as it sucks in oh so many ways :confused:  It's not like they _HAD TO_ make `add_library` not scriptable.

Parsing messages sucks too. Here's an even better overkill:
- find Qt library
- configure template cpp file with `@@vars@@`
- compile
- run
- print JSON with vars to stdout

... and we'll sure win an award for the most bikeshedded open source project. xD

ogoffart:
I've implemented a way to look for include paths and library path with env variables in 82c7357
I hope this fits your need, let me know if it doesn't.
So i'm going to close this pull request.
But thanks a lot anyway for providing a pull request and engaging in the discussion.

Issue #140, opened 2021-05-05
Allow linking against QtQuickControls2 (and more?)


tnibler:
It would be very nice and convenient to enable linking against more qt libraries through feature flags in the qttypes crate. Personally, I only have a need for QtQuickControls2 ([change here](https://github.com/woboq/qmetaobject-rs/commit/308836a20f3149e3e0032762b7a759cb39712b81)), but maybe it would be good to add more libraries as well?

ogoffart:
Yes, that makes total sense!

tnibler:
From https://doc.qt.io/qt-5/qtmodules.html, the libraries we could link against are:

 - `QtQuickControls2`
 - `QtWidgets`
 - `QtMultimedia` and `QtMultimediaWidgets`
 - `QtSql`
 - `QtTest`

Should I just open a PR with all of them?

ratijas:
What would programmers do with them without at least some basic wrappers and support beyond linking?

tnibler:
I guess most of them would only be useful if you want to reuse existing Qt code with some `cpp!` macro glue.

PR #142, opened 2021-05-16
Allow linking against other Qt libraries

tnibler:
This change adds feature flags for the following Qt libraries:

 - QtQuickControls2
 - QtMultimedia
 - QtMultimediaWidgets
 - QtSql
 - QtTest

See #140


ogoffart:
Thank you!

PR #143, opened 2021-05-21
Reduce risk of library mixing when linking against Qt on Linux

tronical:
Similar to macOS, also use rpath on Linux, to encode the location of
where we found Qt.

cc sixtyfpsui/sixtyfps#232

### DIFF
--- a/qttypes/build.rs
+++ b/qttypes/build.rs
@@ -150,7 +150,7 @@ fn main() {
         ""
     };
 
-    if cfg!(target_os = "macos") {
+    if cfg!(any(target_os = "macos", target_os = "linux")) {
         println!("cargo:rustc-cdylib-link-arg=-Wl,-rpath,{}", qt_library_path.trim());
     }
 
--- a/qttypes/build.rs
+++ b/qttypes/build.rs
@@ -150,7 +150,7 @@ fn main() {
         ""
     };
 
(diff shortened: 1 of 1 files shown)


tronical:
For packaging by Linux distros there might eventually come a request to disable rpath again though. But I think for everyone else rpath is good :-) (my biased opinion)

ogoffart:
When i was a Qt developer, i did not want application to set rpath, because that meant it was harder for me to try out different versions or build of the library.
Also this reduce the "movability" of the file, since it would contains path that depends on the system.

But given that it helps for people just wanting to try out the build, it might make sense to apply this patch anyway.

tronical:
Yes, two very valid points. Especially support for $ORIGIN in the rpath would be nice to have.

qttypes/build.rs
```diff
@@ -150,7 +150,7 @@ fn main() {
         ""
     };
 
-    if cfg!(target_os = "macos") {
+    if cfg!(any(target_os = "macos", target_os = "linux")) {
```
ogoffart, line 153: actually, i think the target_os is the target on which the build script is compiled,  not the target which the resulting binary should be built for.

And this will not be great for cross compiling either.

tronical: Indeed, I'll add a commit to this merge request to use an environment variable (I'm sure cargo provides one, will look up :)

PR #144, opened 2021-05-21
Introduce prelude

### LINKED ISSUE #83: Proposal: use prelude::*
This is the proposal draft to include a `prelude` module into `qmetaobject` crate, similar to the one in Rust standard library and many other existing crates. The purpose of the module shall be to re-export all or most of public items from the crate.
Questions to consider:
 - [ ] should all items be exported or only subset of them?
 - [ ] if not all, then how to define a subset?

certainly relates to #81

rubdos:
Fixes #83, with the motto "a semi-good prelude is better than none".

They only "breaking" change (when changing `qmetaobject::*` for `qmetaobject::prelude::*`) would be that I've left out `connect`, since that's quite a generic name. I had to change this at one single place in a doc comment.

I'm still going to try whether I can yeet `lazy_static` too, give me a second here :-)

### DIFF
--- a/qmetaobject/src/lib.rs
+++ b/qmetaobject/src/lib.rs
@@ -22,7 +22,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
     ```
     use cstr::cstr;
-    use qmetaobject::*;
+    use qmetaobject::prelude::*;
 
     // The `QObject` custom derive macro allows to expose a class to Qt and QML
     #[derive(QObject,Default)]
@@ -114,7 +114,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
     This can be done like so:
 
     ```
-    use qmetaobject::*;
+    use qmetaobject::prelude::*;
     # use std::cell::RefCell;
 
     #[derive(QObject, Default)]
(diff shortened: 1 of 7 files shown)


qmetaobject/src/lib.rs
```diff
@@ -209,6 +209,16 @@ pub mod tablemodel;
 #[cfg(feature = "webengine")]
 pub mod webengine;
 
+/// Module intended for glob import.
+pub mod prelude {
+    pub use crate::{
+        lazy_static, qmetaobject_lazy_static, qml_register_type, qrc, qt_base_class, qt_method,
```
ogoffart, line 215: qmetaobject_lazy_static shouldn't be necessary anymore with the changes i did yesterday.

Also, why is lazy_static in there?

qmetaobject/src/lib.rs
```diff
@@ -209,6 +209,16 @@ pub mod tablemodel;
...
+        qt_plugin, qt_property, qt_signal, queued_callback, QAbstractListModel, QByteArray, QColor,
```
ogoffart, line 216: Maybe queued_callback shouldn't be in there either.

ogoffart (commented):
Thanks a lot!

Looks sensible appart from the comments

qmetaobject/src/lib.rs
rubdos: ... oh. I was probably working on an oldish version then. Awesome!

ogoffart:
It looks like i need to add #crate_::  before the use of qmetatype_interface_ptr  (in Qt6)

qmetaobject/src/lib.rs
rubdos: FYI, `lazy_static` is in there, because the `qmeta`-reexport actually doesn't 100% work. `qmetaobject;:*` always pulled `lazy_static` in "by accident" which made it work:

image

I just tried without `qmetaobject_lazy_static`, which still gives me errors on `lazy_static` not being present.

rubdos: Found it, qmeta needs a $crate prefix in its body too :-)

rubdos:
> crate_

Done that in here if you want.

ogoffart:
Thanks!

Note that ideally, the QObject and other macro should work without even importing the prelude.

PR #145, opened 2021-05-24
Additional QVariant conversions

farmaazon:
When I developed my application, I found the conversion from QPointer to QVariant necessary to make an AbstractListModel of QObjects ([an example][https://github.com/farmaazon/dombreaker-rs/blob/master/src/qt_ui/model.rs#L43]).

### DIFF
--- a/qmetaobject/src/lib.rs
+++ b/qmetaobject/src/lib.rs
@@ -532,12 +532,27 @@ impl<'pin, T: QObject + 'pin> QObjectPinned<'pin, T> {
     }
 }
 
+impl<'pin, T: QObject + 'pin> From<QObjectPinned<'pin, T>> for QVariant {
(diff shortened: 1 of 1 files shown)


ratijas:
Hi, thanks for your patch!

That `impl From<QPointF> for QVariant` however is not relevant to QPointer conversion in question, and also it would be a part of an upcoming #136 soon™ anyway. Do you think you could you remove it and force-push amended commit, please?

farmaazon:
> Hi, thanks for your patch!
> 
> That `impl From<QPointF> for QVariant` however is not relevant to QPointer conversion in question, and also it would be a part of an upcoming #136 soon™ anyway. Do you think you could you remove it and force-push amended commit, please?

No problem.

PR #148, opened 2021-05-30
Add `release_resources` to QQuickItem

AdrianEddy:
Add `release_resources` function to `QQuickItem` to allow implementing it on the Rust side.


ogoffart:
Thank you!

PR #152, opened 2021-06-23
Bump some rusty dependencies

ratijas:
Bump dependencies:
- cpp & cpp_build from 0.5.4 to 0.5.6
- semver from 0.11 to 1 (another one reached 1.0, yay!)

### DIFF
--- a/examples/graph/Cargo.toml
+++ b/examples/graph/Cargo.toml
@@ -9,7 +9,7 @@ build = "build.rs"
 qmetaobject = { path = "../../qmetaobject"}
 qttypes = { path = "../../qttypes"} # required to get the DEP_QT_* env variables
 cstr = "0.2"
-cpp = "0.5"
+cpp = "0.5.6"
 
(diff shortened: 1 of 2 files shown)


ogoffart (approved):

ratijas:
I can't believe such CI failure could have been caused by cpp & cpp_build upgrade. Investigating…

ogoffart:
The CI failure is caused by https://github.com/rust-lang/cargo/issues/9554

ratijas:
Shtability of Rusht, huh…

CI can't even pass the stable toolchain's tests just because one of the nightlies failed first. What to do now?  Wait until they nerf it down to a warning?

ratijas:
I wonder, if there's an easy way to temporarily skip checks for nightly, without breaking or somehow altering CI configuration? It's not like we can do anything about the error at this point.

ogoffart:
We can just ignore CI for this one

PR #155, opened 2021-06-27
Fix non-MSVC builds on Windows and cross-compilation in general

### LINKED ISSUE #154: qttypes: Support C++ toolchains on Windows other than MSVC
## Summary

Build.rs script of qttypes assumes that on Windows the toolchain is always MSVC, which leads to linkage errors with e.g. MinGW toolchain.

## Description

As discussed in #150, there are lines in _qttypes/build.rs_ which only work for MSVC setup:

https://github.com/woboq/qmetaobject-rs/blob/d3df3a9de860b77ae2a824ec60c8b94eaffb4c6d/qttypes/build.rs#L146-L151

It produces this output:

```
cargo:rustc-link-lib=msvcrtd
cargo:rustc-link-search=C:/Qt/5.15.2/mingw81_32/lib
cargo:rustc-link-lib=Qt5Cored
cargo:rustc-link-lib=Qt5Guid
cargo:rustc-link-lib=Qt5Widgetsd
cargo:rustc-link-lib=Qt5Quickd
cargo:rustc-link-lib=Qt5Qmld
```

which instructs cargo to pass to a linked a set of libs specifically built for debugging with MSVC toolchain. Libraries with a `d` suffix are only shipped with Qt MSVC builds.

Note that since there is an `if debug` variable check involved in condition, this bug does not happen on `--release` mode.

Also, I don't think `msvcrt`/`msvcrtd` is really needed outside of MSVC ecosystem? Sounds like a compiler's runtime library, similar to [`compiler-rt`](https://compiler-rt.llvm.org/) in LLVM land.

## Steps to reproduce

1. Use Windows
2. Do NOT install Visual Studio / MSVC toolchain (or at least hide it from env for now)
3. Install rustup, add GNU target, say `x86_64-pc-windows-gnu` or `i686-pc-windows-gnu`. To avoid passing around `--target` arguments, default target can be saved in [Cargo configuration](https://doc.rust-lang.org/cargo/reference/config.html).
4. Install Qt with online installer. Make sure to add Qt libs and compiler toolchain built with and for MinGW, matching bitness or Cargo target.
5. Cargo-build qmetaobject-rs itself or as a part of another project in debug mode (i.e. NOT `--release`).

## Expected result

Everything should be OK.

##  Actual result

```
  = note: C:/Qt/Tools/mingw810_32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lmsvcrtd
          C:/Qt/Tools/mingw810_32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lQt5Cored
          C:/Qt/Tools/mingw810_32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lQt5Guid
          C:/Qt/Tools/mingw810_32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lQt5Widgetsd
          C:/Qt/Tools/mingw810_32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lQt5Quickd
          C:/Qt/Tools/mingw810_32/bin/../lib/gcc/i686-w64-mingw32/8.1.0/../../../../i686-w64-mingw32/bin/ld.exe: cannot find -lQt5Qmld
          collect2.exe: error: ld returned 1 exit status
```

## Now what?

Add to qttypes build script a check for current toolchain being used, and only add compiler runtime and `d` suffix if target is related to MSCV.

## Version

My best guess is qmetaobject-rs starting from v0.2. It doesn't happen on published v0.1.4, i.e. before factoring out qttypes into its own crate.

ratijas:
Simple `cfg!(target_* = "...")` doesn't work in build scripts the way it
does in crate's code, because build scripts are being compiled for the
toolchain's target triple -- not the default-host nor the cargo's
eventual target. Instead, such configuration should be loaded from
provided environment variables at run-time.

While initially trying to fix linkage issues for Windows + MinGW target
(first reported by @stuartZhang in #150), I came to realize in the
process that `cfg!(target_os = ...)`, although being used only for OS X
at this moment, is likely screwed up in the same way (except, who would
ever cross-compile to or from Mac, and why?)

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

Fixes #154

---

My life priorities are seriously questionable, and it almost makes me wonder why did I choose this job in the first place.  Now to get some well-deserved rest...

### DIFF
--- a/qttypes/build.rs
+++ b/qttypes/build.rs
@@ -30,10 +30,12 @@ fn qmake_query(var: &str) -> Result<String, std::io::Error> {
 }
 
 fn open_header(file: &str, qt_include_path: &str, qt_library_path: &str) -> std::fs::File {
+    let cargo_target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+
     let mut path = PathBuf::from(qt_include_path);
     path.push("QtCore");
     path.push(file);
-    if cfg!(target_os = "macos") {
+    if cargo_target_os == "macos" {
         if !path.exists() {
             path = Path::new(qt_library_path).join("QtCore.framework/Headers");
             path.push(file);
@@ -77,6 +79,11 @@ fn detect_version_from_header(qt_include_path: &str, qt_library_path: &str) -> S
 }
 
 fn main() {
+    // Simple cfg!(target_* = "...") doesn't work in build scripts the way it does in crate's code.
+    // https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts
+    let cargo_target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+    let cargo_target_env = std::env::var("CARGO_CFG_TARGET_ENV").unwrap();
+
     println!("cargo:rerun-if-env-changed=QT_INCLUDE_PATH");
     println!("cargo:rerun-if-env-changed=QT_LIBRARY_PATH");
     let (qt_version, qt_include_path, qt_library_path) = match (
@@ -117,7 +124,7 @@ fn main() {
 
     let mut config = cpp_build::Config::new();
 
-    if cfg!(target_os = "macos") {
+    if cargo_target_os == "macos" {
         config.flag("-F");
         config.flag(qt_library_path.trim());
     }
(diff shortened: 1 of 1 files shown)


ogoffart (approved):
Looks good, thanks

PR #166, opened 2021-07-03
Yet another cleanup round

ratijas:
A bunch of clean-ups including but not limited to code style cosmits. Not much, but it's been lying around in my stash and bugging me out for too long.

* Fix bare_trait_objects deprecation warning
* example(graph): Move main.qml into qrc
* cosmit: Use modern ECMAScript `let` instead of var
* cosmit: Unify cpp! {{ ... }} macro style
* Wrap test code in `#[cfg(test)] mod tests` boilerplate
* cosmit: Prettify scenegraph macro code
* cosmit: Add visual spacing to qrc! docs

### DIFF
--- a/qmetaobject_impl/src/lib.rs
+++ b/qmetaobject_impl/src/lib.rs
@@ -89,7 +89,7 @@ pub fn qenum_impl6(input: TokenStream) -> TokenStream {
     qobject_impl::generate_enum(input, 6)
 }
 
-/// Implementation of the qmetaobject::qrc! macro
+// Implementation of the qmetaobject::qrc! macro
 #[proc_macro]
 pub fn qrc_internal(input: TokenStream) -> TokenStream {
     qrc_impl::process_qrc(input)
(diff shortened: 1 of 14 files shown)


qmetaobject_impl/src/lib.rs
```diff
@@ -89,7 +89,7 @@ pub fn qenum_impl6(input: TokenStream) -> TokenStream {
     qobject_impl::generate_enum(input, 6)
 }
 
-/// Implementation of the qmetaobject::qrc! macro
+// Implementation of the qmetaobject::qrc! macro
```
ogoffart, line 92: why remove this documentation?

ogoffart (approved):
Looks ok.

qmetaobject_impl/src/lib.rs
ratijas: Because it showed up at the bottom of generated docs. And that wasn't pretty.

ratijas: It stacks up with documentation on the re-export in the user-facing `qmetaobject` crate.

ratijas: For the reference, here it is at the bottom: https://docs.rs/qmetaobject/0.2.2/qmetaobject/macro.qrc.html

ogoffart: ah ok.

Issue #167, opened 2021-07-03
Disable WebEngine example on Windows + MinGW


ratijas:
## Summary

Qt does not ship pre-built WebEngine component for Windows MinGW target, so we should find a way to exclude it from the list of default targets on that platform's target triple.

## Description

Running simple `cargo build` in qmetaobject-rs repository on Windows with MinGW environment results in build failure:

```
   Compiling qmetaobject v0.2.2 (C:\Users\ratijas\projects\qmetaobject-rs\qmetaobject)
   Compiling graph v0.1.0 (C:\Users\ratijas\projects\qmetaobject-rs\examples\graph)
The following warnings were emitted during compilation:

warning: src\webengine.rs:4:10: fatal error: QtWebEngine/QtWebEngine: No such file or directory
warning:  #include <QtWebEngine/QtWebEngine>
warning:           ^~~~~~~~~~~~~~~~~~~~~~~~~
warning: compilation terminated.

error: failed to run custom build command for `qmetaobject v0.2.2 (C:\Users\ratijas\projects\qmetaobject-rs\qmetaobject)`

Caused by:
  process didn't exit successfully: `C:\Users\ratijas\projects\qmetaobject-rs\target\debug\build\qmetaobject-4c86368b25d7a0a5\build-script-build` (exit code: 1)
  --- stdout
  TARGET = Some("i686-pc-windows-gnu")
  OPT_LEVEL = Some("0")
  HOST = Some("x86_64-pc-windows-gnu")
[...]
  running: "g++.exe" "-O0" "-ffunction-sections" "-fdata-sections" "-g" "-fno-omit-frame-pointer" "-m32" "-I" "C:\\Users\\ratijas\\projects\\qmetaobject-rs\\qmetaobject" "-I" "C:/Qt/5.15.2/mingw81_32/include" "-Wall" "-Wextra" "-std=c++11" "-o" "C:\\Users\\ratijas\\projects\\qmetaobject-rs\\target\\i686-pc-windows-gnu\\debug\\build\\qmetaobject-54d9dbd8cb0dbb47\\out\\rust_cpp\\cpp_closures.o" "-c" "C:\\Users\\ratijas\\projects\\qmetaobject-rs\\target\\i686-pc-windows-gnu\\debug\\build\\qmetaobject-54d9dbd8cb0dbb47\\out\\rust_cpp\\cpp_closures.cpp"
[...]

PR #168, opened 2021-07-03
Silence deprecation warning in Graph example

ratijas:
```
           Such solution

much compatibility
                               very fix

                      wow
```

After spending a day trying to figure out this new shiny undocumented
experimental Qt Rhi, I finally gave up and implemented a workaround
that disables compiler's deprecation warnings altogether. Those old
QSGSimpleMaterial{,Shader} classes ain't going to be removed from Qt5
which is on a life support at this point; and we know for sure they are
already gone in Qt6. So, there's just no point seeing these warning
over and over again.


ogoffart:
Maybe add a comment in the code to explain why we disable these warnings.

(Since this example isn't working with Qt6, maybe it should be disabled when building with Qt6)

ratijas:
> Maybe add a comment in the code to explain why we disable these warnings.

Might be a good idea, since not everyone likes to read git history.

> (Since this example isn't working with Qt6, maybe it should be disabled when building with Qt6)

I have no idea how to implement such feature-based workspace members exclusion. Also, reference to #167.

ogoffart:
> I have no idea how to implement such feature-based workspace members exclusion.

Something like 

```
if std::env::var("DEP_QT_VERSION").unwrap() > "6." { /* don't compile this */ }
```

ratijas:
> `/* don't compile this */`

Yeah, but, like, how? Provide it with a dumb stub? Make it look like it actually compiles?

Or maybe just disable examples by default, and adjust CI configuration.

ogoffart:
Using a feature?
I do the same with qttypes so it still builds even if no qt is found.

https://github.com/woboq/qmetaobject-rs/blob/757288a026c9fd9a62ad84e54596c4d07423bff8/qttypes/build.rs#L112-L112

ratijas:
>  I do the same with qttypes so it still builds even if no qt is found.

Ah yes, I remember now. It builds into some unusable shape just for Rust type-checking and documentation purposes. That might be a good way to solve both this and #167.

ratijas:
That's a lot of changes, so I created a branch in the middle, right where I stopped improving _build.rs_, and started actually fixing problems.

ratijas:
Done with rebasing and force-pushing.

PR #172, opened 2021-07-10
Refactor RustObjectDescription

ratijas:
Renames it to ...Descriptor. Both are nouns, but description is
something like a documentation text, while descriptor is something
item-like that can be owned, passed around etc. Besides, its a popular
naming choice, e.g.: file descriptor, process descriptor etc.

Adds letter Q to the name to emphasize the fact that it specifically
deals with QObjects, not just any object wrapped in Rust structs.

Converts rustObjectDescription() stand-alone function into single
provider static method RustQObjectDescriptor::instance(). It's
convenient that it is namespaced, and generally looks good.

---

It's almost like I have nothing better to do with my life at this point.

On the bright side, I finally learned how to implement custom QObject subclasses.
Gotta write structured tutorial someday.


ogoffart (dismissed):
Make sense.
But renaming QObjectDescription might break semver.
Maybe you can add a deprectated type alias:

```rust
#[doc(hidden)] #[deprecated] pub type  QObjectDescription = QObjectDescriptor
```

ratijas:
Oh sh~… Thanks, nice catch! Even though it was `#[doc(hidden)]`, it's still a part of a public API.

ratijas:
If that was the only concert, I guess requesting extra review a redundant waste of time.

PR #174, opened 2021-07-10
Shiny build scripts improvements

ratijas:
Actually, just a bunch of cleanups.

---

 * build.rs: Uplift stdout trimming into qmake_query
 * build.rs: Uplift BufReader of headers, and other minor things
 * build.rs: Reorder worst-case scenario last
 * build.rs: Use semver crate in qttypes
 * cosmit: alphabetic sort and spacing for Cargo.toml


ogoffart:
Thank you, many great changes. I saw this PR after the qtwebengine one, so i left some comment in the other PR.

The only think i don't really like is the added dependencies to semver.

ratijas:
> The only think i don't really like is the added dependencies to semver.

I'd say, it's not added, but rather now shared with other crates. And uniform handling of versions definitely improves clarity. I don't think it's a big deal, especially because it was already there since ages. And it doesn't have dependencies on its own.

ratijas:
Fixed horrible typo in commit message.

ogoffart (approved):

ratijas:
Rebased onto master, resolved merge conflict, reverted `&format()` in a loop to a plain text message — might worth refactoring into a function scanning over header's lines though.

ogoffart (approved):

ratijas:
Created rollup merge commits, just for fun, to play around with multi-parent octopus strategy. Since we don't have separate staging/rollup branch, I pushed them to ratijas:master (because who cares about ratijas:master). Waiting for CI there, then forwarding to upstream.

ratijas:
Oh, it's done already: https://github.com/ratijas/qmetaobject-rs/actions/runs/1025994331

PR #175, opened 2021-07-10
Disable WebEngine on non-supported platforms and Qt versions

### LINKED ISSUE #167: Disable WebEngine example on Windows + MinGW

ratijas:
To be merged after #174.  Independent of #168, but deals with the same kind of problems.

Disable WebEngine under certain conditions that ruined `cargo build` from the root of the workspace:
- on Qt >= 6.0 && < 6.2
- for Windows non-MSVC targets (namely, MinGW)

Fixes #167


qttypes/build.rs
```diff
@@ -64,12 +64,12 @@ fn detect_qreal_size(qt_include_path: &str, qt_library_path: &str) {
...
 fn detect_version_from_header(qt_include_path: &str, qt_library_path: &str) -> String {
-    let f = open_header("qtcoreversion.h", qt_include_path, qt_library_path);
-    let b = BufReader::new(f);
+    const VERSION_HEADER: &'static str = "qtcoreversion.h";
+    let b = open_core_header(VERSION_HEADER, qt_include_path, qt_library_path);
 
     // Find declaration of QTCORE_VERSION_STR
     for line in b.lines() {
-        let line = line.expect("qtcoreversion.h is valid UTF-8");
+        let line = line.expect(&format!("UTF-8 conversion failed for `{:?}`", VERSION_HEADER));
```
ogoffart, line 74: nitpick:  format in expect is ineficcient because the allocation and string formating will happen no matter what. 
maybe the best instead is to use a `.or_else(|| panic!(...))`  construct

qttypes/Cargo.toml
```diff
@@ -40,6 +40,7 @@ chrono = { version = "0.4", optional = true }
 
 [build-dependencies]
 cpp_build = "0.5.6"
+semver = "1"
```
ogoffart, line 43: i'm a bit disappointed by adding this dependencies just for detecting Qt6.  it's a bit overkill and will not help with compile time.

ogoffart (approved):
note that longterm, i don't think it make sense to support 6.0 and 6.1

qttypes/build.rs
ratijas: Yeah, I didn't notice it's in the loop. I thought, like, who cares about build script micro-optimizations. But this is exactly not the case.

qttypes/Cargo.toml
ratijas: I checked it, and it's 80K of Rust files, mostly doc comments, serde support and tests (OMG so many tests). After stripping it down, it's about 30K of code, no generics. Still a bit of an overkill, especially comparing to C++ `#define QT_VERSION_CHECK()`, but oh well… that's what rust it all about :)

qttypes/build.rs
ratijas: Reverted to a plain `expect` with a message like in this `format`.

PR #181, opened 2021-07-20
Clean up and document QRustClosureSlotObject implementation

ratijas:
Moves Q_UNUSED variables to the top of the function body.

Renames generated Rust closures (the ones inside rust! macro) to match
outer C++ class name.

Enhances docs on what's going on there.

Generally, this is like a second take on #104, but without actually
attempting to remove NumOperations enum variant.


ratijas:
Wait a sec. Docs on `ret` is incorrect. That's a return flag for comparison (if implemented).

ogoffart (approved):

PR #182, opened 2021-07-21
Refactor write_i32 and string data builder

ratijas:
Reduces cognitive load by some amount, and introduces more SAFETY
comments -- sometimes they are needed even in absence of unsafe
{ blocks }, like with narrowing integral types casts.

---

Yet another long-forgotten stashed change from my local copy

### DIFF
--- a/qmetaobject_impl/src/qobject_impl.rs
+++ b/qmetaobject_impl/src/qobject_impl.rs
@@ -475,9 +475,9 @@ pub fn generate(input: TokenStream, is_qobject: bool, qt_version: QtVersion) -> 
                             let mut flags = 1 | 2 | 0x00004000 | 0x00001000 | 0x00010000;
                             for it in parsed.1 {
                                 match it {
-                                    Flag::Notify(signal) => {
+                                    Flag::Notify(i) => {
                                         assert!(notify_signal.is_none(), "Duplicate NOTIFY for a property");
-                                        notify_signal = Some(signal);
+                                        notify_signal = Some(i);
                                         flags |= 0x00400000;
                                     }
                                     Flag::Const => {
--- a/qmetaobject_impl/src/qobject_impl.rs
+++ b/qmetaobject_impl/src/qobject_impl.rs
@@ -95,13 +95,8 @@ impl IsVoid for syn::Type {
     }
 }
 
-fn write_u32(val: i32) -> [u8; 4] {
-    [
-        (val & 0xff) as u8,
-        ((val >> 8) & 0xff) as u8,
-        ((val >> 16) & 0xff) as u8,
-        ((val >> 24) & 0xff) as u8,
-    ]
+fn write_i32(vec: &mut Vec<u8>, val: i32) {
+    vec.extend_from_slice(&val.to_le_bytes())
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


qmetaobject_impl/src/qobject_impl.rs
```diff
@@ -160,37 +155,45 @@ impl MetaObject {
...
 
         if self.qt_version == 5 {
             let sizeof_qbytearraydata = if target_pointer_width == 64 { 24 } else { 16 };
             let mut ofs = sizeof_qbytearraydata * self.string_data.len() as i32;
             for s in self.string_data.iter() {
-                result.extend_from_slice(&write_u32(-1)); // ref (-1)
-                result.extend_from_slice(&write_u32(s.len() as i32)); // size
-                result.extend_from_slice(&write_u32(0)); // alloc / capacityReserved
+                // Make sure the string fits even with trailing null ('\0').
+                assert!(s.len() + 1 <= i32::MAX as usize, "string data is too large");
```
ogoffart, line 165:
```suggestion
                assert!(s.len() <= i32::MAX as usize, "string data is too large");
```

To avoid all possibility of overflow
If it is so big it would also not work anyway

ogoffart (approved):

qmetaobject_impl/src/qobject_impl.rs
ratijas: I'm not even sure I wanna test that 🥲

But wait, why wouldn't it work with `s.len() == i32::MAX - 1`? We have to use some length check anyway, so why removing `+ 1` should help?

ratijas: I mean, `i32::MAX` is just 4GB of data which I could test locally alright, but then we'll surely get murdered by CI.

ogoffart: I think the internal of QByteArray don't allow sizes that are bigger than 31 bits

ratijas:
Rebased and reordered, such that most important commit goes last, hence stays on top.

qmetaobject_impl/src/qobject_impl.rs
ratijas: Yeah, it was written [in the docs](https://doc.qt.io/qt-5/qbytearray.html#maximum-size-and-out-of-memory-conditions) all this time.

> The current version of QByteArray is limited to just under 2 GB (2^31 bytes) in size. The exact value is architecture-dependent, since it depends on the overhead required for managing the data block, but is no more than 32 bytes. Raw data blocks are also limited by the use of int type in the current version to 2 GB minus 1 byte.

The wording is a little bit confusing as to what it means: "The exact value is architecture-dependent […] but is no more than 32 bytes". Does this mean that limit is 2^31 minus 32-or-less bytes? [Raw] data blocks are also not mentioned anywhere else on the page. Couldn't they just provide a compile-time constant like MAX_SIZE?

It's hard to trace source code with all these layers of internals involved. I wasn't able to find any size checks there.

Either way, I tried in C++ Qt console app, and I was able to append 2^32 = 4294967296 bytes to QByteArray no problems. Both Qt 5.15.2 and 6.1.2 64bit. Maybe docs are outdated w.r.t. current capabilities, or it's just 64bit architecture?

Oh well, whatever. Let's just check for casts overflow and nothing else.

ratijas:
Casts safety guarantees. Finally.

So, basically I pulled up checks on string data's vector and per-string length; and wrapped all math in `checked_*().unwrap()` calls. Error reporting isn't perfect, but that's a topic for another PR.

ratijas:
~~Oh yeah, labels.~~

I'd like you to shoot another glance at it, sir. Just to be on a safe side.

ogoffart (approved):

PR #185, opened 2021-07-21
Add wrappers around public static getters & setters of QCoreApplication

### LINKED ISSUE #113: Expose QCoreApplication properties for Dialogs
The Dialogs QML module uses QSettings on some platforms for things like window dimensions and remembering previous locations but doesn't work without set values on the qApp for 'organizationName', 'organizationDomain', and 'applicationName'.

> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: Failed to initialize QSettings instance. Status code is: 1
> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: The following application identifiers have not been set: QVector("organizationName", "organizationDomain", "applicationName")
> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: Failed to initialize QSettings instance. Status code is: 1
> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: The following application identifiers have not been set: QVector("organizationName", "organizationDomain", "applicationName")
> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: Failed to initialize QSettings instance. Status code is: 1
> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: The following application identifiers have not been set: QVector("organizationName", "organizationDomain", "applicationName")
> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: Failed to initialize QSettings instance. Status code is: 1
> 19:35:49 [WARN] file:///usr/lib/qt/qml/QtQuick/Dialogs/DefaultFileDialog.qml:102:33: QML Settings: The following application identifiers have not been set: QVector("organizationName", "organizationDomain", "applicationName")

Having methods to set these, maybe even inferring applicationName from the `CARGO_PKG_NAME` would make these modules work.

This would also let the qt.labs.settings module to work as well.

ratijas:
To be honest, I'm a bit concerned about the name of this new module, because core is also a part of rustc std.

---

Closes #113

### DIFF
--- a/qmetaobject/src/lib.rs
+++ b/qmetaobject/src/lib.rs
@@ -195,6 +195,7 @@ pub use qttypes::*;
 pub use tablemodel::*;
 
 pub mod connections;
+pub mod qtcore;
 pub mod future;
 pub mod itemmodel;
 pub mod listmodel;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 - Tutorial on adding Rust wrappers #171.
+- QCoreApplication: wrappers around public static getters & setters #185.
 
 ### Deprecated
 - Rename QObjectDescription in favor of its new name RustQObjectDescriptor #172.
--- a/qmetaobject/src/qmetatype.rs
+++ b/qmetaobject/src/qmetatype.rs
@@ -138,7 +138,7 @@ fn register_metatype_common<T: QMetaType>(
         let deleter_fn: extern "C" fn(v: *mut T) = deleter_fn;
 
         #[cfg(qt_6_0)]
-        let deleter_fn: *const c_void = core::ptr::null();
+        let deleter_fn: *const c_void = ::core::ptr::null();
 
         #[cfg(not(qt_6_0))]
         extern "C" fn creator_fn<T: Default + Clone>(c: *const T) -> *const T {
@@ -240,7 +240,7 @@ fn register_metatype_common<T: QMetaType>(
                 alignment: u16,
                 size: u32,
                 flags: u32,
-                typeId: core::sync::atomic::AtomicI32,
+                typeId: ::core::sync::atomic::AtomicI32,
                 metaObjectFn: extern "C" fn(*const QMetaTypeInterface)-> *const QMetaObject,
                 name: *const c_char,
                 defaultCtr:  extern "C" fn(*const QMetaTypeInterface, *mut c_void),
--- a/qmetaobject/src/qtcore/core_application.rs
+++ b/qmetaobject/src/qtcore/core_application.rs
@@ -1,0 +1,132 @@
+/* Copyright (C) 2018 Olivier Goffart <ogoffart@woboq.com>
+   Copyright (C) 2021 ivan tkachenko a.k.a. ratijas <me@ratijas.tk>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+associated documentation files (the "Software"), to deal in the Software without restriction,
+including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial
+portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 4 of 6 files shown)


qmetaobject/src/core.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn application_name() -> String {
+        cpp!(unsafe [] -> QString as "QString" {
+            return QCoreApplication::applicationName();
+        }).to_string()
+    }
+
+    /// Wrapper around [`void setApplicationName(const QString &application)`][qt] static method.
+    ///
+    /// [qt]: https://doc.qt.io/qt-5/qcoreapplication.html#applicationName-prop
+    pub fn set_application_name(application: &str) {
```
ogoffart, line 57: Why not take a QString since we take ownership as a QString?

ratijas: AFAIK String to QString conversion is not in-place, and re-allocated memory. So, why drop a String if we only need a str slice? This follows minimal requirements rule.

Here are the conversion impls, btw
```rust
impl<'a> From<&'a str> for QString {
    /// Copy the data from a `&str`.
    fn from(s: &'a str) -> QString {
        let len = s.len();
        let ptr = s.as_ptr();
        cpp!(unsafe [len as "size_t", ptr as "char*"] -> QString as "QString" {
            return QString::fromUtf8(ptr, len);
        })
    }
}
impl From<String> for QString {
    fn from(s: String) -> QString {
        QString::from(&*s)
    }
}
```

ratijas: Hilariously, documentation of [QString::fromUtf8](https://doc.qt.io/qt-5/qstring.html#fromUtf8) says nothing about ownership. It just "returns a QString initialized with the first size bytes of the UTF-8 string str" — whatever that "initialization" might imply. And the type of `const char *str` requiring it being const doesn't make the whole situation better. Here we go reading source code once again…

ratijas: I think, the broad question here is: do we want to make wrappers more "Rust idiomatic" by exposing QString as String/&str etc., or do we want preserve the implicit sharing provided by Qt data types (not only strings)?

ogoffart: Right, QString is owning the string.    So it allocates.  Just like `String` in rust.

> So, why drop a String if we only need a str slice?

We don't need a str slice,  we need a QString to give it to the application.

ratijas: Shout-outs to #81, by the way. I think it was mentioned somewhere there, but was never resolved. If not, then it should be added.

So, give me a second to refactor it into QStrings.

ratijas:
I assume, this is good to merge then?

Free (like in beer) weekends are coming, and I hope to start hacking on new stuff, or rather splitting the old good macro into parser/generator steps.

qmetaobject/src/lib.rs
```diff
@@ -195,6 +195,7 @@ pub use qttypes::*;
 pub use tablemodel::*;
 
 pub mod connections;
+pub mod core;
```
ogoffart, line 198: not sure core is a good name for a module because of the core crate.   Maybe core_application or just application would be better.

ogoffart (approved):
:+1: 
Yes, good to merge.
Althouhg maybe you want to rename the `core` module

qmetaobject/src/lib.rs
ratijas: So one Rust module per "class" then? And what about Qt modules, i.e. QtCore?  I recall we never reached consensus on this topic either. I believe it is worth experimenting with both styles here and there, and see which one works best in the end.

How about _qtcore_ for now, and maybe refactor into _qtcore::core_application_ later if needed? At the very least, that would go in-line with _qtdeclarative_ and _qtquickcontrols2_ modules.

ogoffart: yeah, i don't know if it makes sense to have one rust module per Qt library.  
maybe per group of class that belong together.

ratijas: It is hard, and probably meaningless, to try to define a group of related classes. That would be some opinionated and not well-known mapping which would need to be documented somewhere.

On the other hand, file per class looks reasonable, because (1) that's what Qt already does, and (2) classes have so called "related non-members" and associated enums that would fit in such rust file/module just fine.

If there is ever going to be too many wrappers, at that point it might be feasible to split into "qmetaobject_qt{core,declarative,…}" crates to take advantage of parallel compilation in multi-threaded environments.

ratijas:
Refactored into _qtcore::core_application_ module. Added a note to clarify that `quit()` method is not and should not have `!` never type, because it actually returns.

> Free (like in beer) weekends are coming, and I hope to start hacking on new stuff, or rather splitting the old good macro into parser/generator steps.

That… didn't age well. All I've really done is [fixed JSON.stringify](https://bugreports.qt.io/browse/QTBUG-95324) in Qt V4 JavaScript engine, learned few new tricks in Zelda Breath of the Wild — and there goes my weekend.

Issue #39, opened 2019-06-19
Registering C-style integer enums?


Ace4896:
How would I register a C-style integer enum e.g.

```rust
enum MyEnum {
    A = 1,
    B = 2,
    // etc
}
```

so that I could use it in both Rust and QML?

Currently, I'm working around this by duplicating the enum definition in my QML and Rust code, where I cast the Rust enum as seen [here](https://stackoverflow.com/a/31359044) and pass around integer values, but is there a way of registering this type in QML directly?

ogoffart:
right, this is not yet implemented.
Not sure how it could work. 
We would need the derive macro to know about the name of each enum value, so that means it probay should be declared inside the struct definition. But rust don't allow that. So it would have to be wrap in a macro or an attribute.

Guiguiprim:
Maybe we could exploit namespace enums (since Qt 5.8 https://www.kdab.com/new-qt-5-8-meta-object-support-namespaces/amp/) to start introducing registered enums.

```rust
#[derive(QEnums)]
enum MyEnum { ... }
```

Could generate a metaobject as if

```cpp
namespace MyEnum {
enum MyEnum { ... };
Q_NAMESPACE
Q_ENUM_NS(MyEnum)
}
```

Guiguiprim:
I'm going to try to add QEnum based on this idea.

I might need to finalize the INT_DATA to create the static QMetaObject during the lazy initialization as I don't think it's possible to get the enums descriminant values during code-gen.

Also I will need to ensure that the property type name match a meta-type name (MyEnum vs MyEnum::MyEnum).

Should be fun!

ogoffart:
Right, so then the enum will be "MyEnum::MyEnum" and there would be one metaobject only for this enum.
I guess that should work.

ratijas:
I tried this code, and it works as of `qmetaobject-rs` v0.1.4:

```rust
use qmetaobject::*;

#[derive(Copy, Clone, Debug, Eq, PartialEq)]
#[derive(QEnum)]
#[repr(u32)]
pub enum Role {
    Default = 1,
    Service = 2,
}

let uri = cstr!("MyTest");
qml_register_enum::<Role>(uri, 1, 0, cstr!("Role"));
```

```qml
import MyTest 1.0
Text {
    text: `${Role.Service}`
}
```

It was implemented by @Guiguiprim back in 2019 via #42. Thus, I'm closing this issue. Thanks, Guiguiprim!

PR #192, opened 2021-08-09
Add QEnum to prelude and add example

sztomi:
I had a hard time figuring this out, but found this comment: https://github.com/woboq/qmetaobject-rs/issues/39#issuecomment-866254747

I think it's worthwhile to add an example that shows how to use it. 

The rationale behind adding qml_register_enum to the prelude is that these free functions are included early in the C++ headers and are expected to be "just there" pretty much always. Additionally, `qml_register_type` is in the prelude so I figured it makes sense to add `qml_register_enum` as well.

### DIFF
--- a/qmetaobject/src/lib.rs
+++ b/qmetaobject/src/lib.rs
@@ -216,9 +216,9 @@ pub mod webengine;
 pub mod prelude {
     pub use crate::{
         qml_register_type, qrc, qt_base_class, qt_method, qt_plugin, qt_property, qt_signal,
-        QAbstractListModel, QByteArray, QColor, QDate, QDateTime, QModelIndex, QObject, QObjectBox,
-        QPointer, QQmlExtensionPlugin, QQuickItem, QQuickView, QRectF, QString, QTime, QVariant,
-        QmlEngine,
+        qtdeclarative::qml_register_enum, QAbstractListModel, QByteArray, QColor, QDate, QDateTime,
+        QEnum, QModelIndex, QObject, QObjectBox, QPointer, QQmlExtensionPlugin, QQuickItem,
+        QQuickView, QRectF, QString, QTime, QVariant, QmlEngine,
     };
 }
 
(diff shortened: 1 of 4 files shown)


qmetaobject/src/lib.rs
```diff
@@ -215,10 +215,32 @@ pub mod webengine;
 /// Module intended for glob import.
 pub mod prelude {
     pub use crate::{
-        qml_register_type, qrc, qt_base_class, qt_method, qt_plugin, qt_property, qt_signal,
-        QAbstractListModel, QByteArray, QColor, QDate, QDateTime, QModelIndex, QObject, QObjectBox,
-        QPointer, QQmlExtensionPlugin, QQuickItem, QQuickView, QRectF, QString, QTime, QVariant,
+        QAbstractListModel,
```
sztomi, line 218: I realize this might be controversial, but it was difficult for me to guess what would be the right location for `QEnum` and `qml_register_enum`. If I added the function after `qt_signal` I would have had to re-wrap all the lines. At the expense of an uglier diff in this PR, this list will be more easier to edit in the future (with diffs only showing added/deleted items).

ogoffart: I believe actually cargo fmt is wrapping and sorting these.

ogoffart (approved):
Thanks for the patch!
I'll merge if CI passes.

qmetaobject/src/lib.rs
sztomi: Oh, my bad then. I saw in another thread fmt is not used in this project so I assumed that this was handcrafted. I can reformat it then.

sztomi: Reformatted it. There was quite a bit more formatting that `cargo fmt` applied but I didn't stage those changes.

ogoffart:
I think something went wrong because now you just removed a bunch of stuff :-/

sztomi:
@ogoffart yep, sorry about that. I got confused between  `git add -p` and stashing my changes. It should be good now!

ogoffart:
Thanks!

PR #196, opened 2021-09-14
Put qml_register_enum behind Qt5.8 guard

### LINKED ISSUE #195: Qt 5.6: prelude import of register_enum
Commit 4772476b02e745110c0b75ae29d86aa0ed2115cc by @sztomi rightfully imports `qml_register_enum` in the prelude, but `qml_register_enum` only gets generated on Qt 5.8+, due to its config directive:

```rust
/// Register the given enum as a QML type.
///
/// Refer to the Qt documentation for [qmlRegisterUncreatableMetaObject][qt].
///
/// [qt]: https://doc.qt.io/qt-5/qqmlengine.html#qmlRegisterUncreatableMetaObject
#[cfg(qt_5_8)]
pub fn qml_register_enum<T: QEnum>(
    uri: &CStr,
    version_major: u32,
    version_minor: u32,
    qml_name: &CStr,
) {
    let uri_ptr = uri.as_ptr();
    let qml_name_ptr = qml_name.as_ptr();
    let meta_object = T::static_meta_object();

    cpp!(unsafe [
            qml_name_ptr as "char *",
            uri_ptr as "char *",
            version_major as "int",
            version_minor as "int",
            meta_object as "const QMetaObject *"
        ] {
    #if QT_VERSION >= QT_VERSION_CHECK(5, 8, 0)
            qmlRegisterUncreatableMetaObject(
                *meta_object,
                uri_ptr,
                version_major,
                version_minor,
                qml_name_ptr,
                "Access to enums & flags only"
            );
    #endif
        })
}
```

Now, there are *two* version guards on this function.  We can choose to drop the `cfg()`-guard, which exposes a no-op `qml_register_enum` via prelude, or we can choose to drop the inner guard, and add a guard on the `prelude` re-export.

I'm not sure which route to take here, but I'll gladly make a PR to fix this.

rubdos:
Fixes #195

### DIFF
--- a/qmetaobject/src/lib.rs
+++ b/qmetaobject/src/lib.rs
@@ -214,11 +214,13 @@ pub mod webengine;
 
 /// Module intended for glob import.
 pub mod prelude {
+    #[cfg(qt_5_8)]
+    pub use crate::qtdeclarative::qml_register_enum;
     pub use crate::{
         qml_register_type, qrc, qt_base_class, qt_method, qt_plugin, qt_property, qt_signal,
-        qtdeclarative::qml_register_enum, QAbstractListModel, QByteArray, QColor, QDate, QDateTime,
-        QEnum, QModelIndex, QObject, QObjectBox, QPointer, QQmlExtensionPlugin, QQuickItem,
-        QQuickView, QRectF, QString, QTime, QVariant, QmlEngine,
+        QAbstractListModel, QByteArray, QColor, QDate, QDateTime, QEnum, QModelIndex, QObject,
+        QObjectBox, QPointer, QQmlExtensionPlugin, QQuickItem, QQuickView, QRectF, QString, QTime,
+        QVariant, QmlEngine,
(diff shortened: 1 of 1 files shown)


rubdos:
Seems to happily build Whisperfish now :-)

ratijas:
It's like the third time already I'm feeling pain from looking at that long list's diff. Might be a good idea to finally split it one item per line (it also might require some cargo-fmt specific attributes to maintain it). For now, I diffed manually in text editor, and everything looks correct.

Anyway, ahem…

Thanks for the patch!

ratijas (approved):
((The comment above should've been this approval, but whatever))

rubdos:
> Might be a good idea to finally split it one item per line (it also might require some cargo-fmt specific attributes to maintain it).

Maybe split it up in several smaller `use` blocks that make sense separately? Then the diffs become a lot clearer already.

PR #199, opened 2021-09-15
Fix compilation with Qt 6.2 and MSVC

tronical:
MSVC does not set the __cplusplus macro correctly by default. It is necessary to pass /Zc:__cplusplus,
which in turn Qt 6.2 requires now.

See also https://codereview.qt-project.org/c/qt/qtbase/+/334623

### DIFF
--- a/qmetaobject/build.rs
+++ b/qmetaobject/build.rs
@@ -36,6 +36,7 @@ fn main() {
     if qt_version >= Version::new(6, 0, 0) {
         config.flag_if_supported("-std=c++17");
         config.flag_if_supported("/std:c++17");
+        config.flag_if_supported("/Zc:__cplusplus");
     }
     config.include(&qt_include_path).build("src/lib.rs");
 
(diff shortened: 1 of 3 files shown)


ratijas:
I assumed I-crash because it breaks the build, right?

tronical:
I don’t know exactly the intended meaning of the I-Crash label but I can confirm that without this change the build aborts.

ogoffart:
I wouldn't call a build breakage a crash. But that doesn't really matter 🙂

ogoffart:
I wonder if flag_if_supported does not slow down the build quite a lot because it does a try with the flag, and then without.

ratijas:
> I wouldn't call a build breakage a crash. But that doesn't really matter slightly_smiling_face

Well, we don't really expect ICE (internal compiler errors) because of qmetaobject crates — that would be too much for us to handle in this repo. So while I just copied labels from rust-lang, I think we can settle on slightly different meaning of this particular one, I-crash. "Compiler or runtime crashes" -> "Build breakage, compiler or runtime crashes". I don't think it's worth having separate label for builds, because in the end it's just some unusable state of a product anyway.

PR #202, opened 2021-10-19
Added set_object method for QQmlEngine.

Ayush1325:
Calls `QQmlEngine::rootContext()->setContextObject` internally.

This is needed for using KI18n with Qt. Might also be useful to work with other KDE frameworks.


ratijas:
KI18n, huh... Just as a was having an idea of adding tr() methods to proc macro codegen...

ogoffart:
thanks for the patch!

PR #204, opened 2021-10-31
fix typo

LastLightSith:


ratijas:
Oops :P

Thanks for fixing it! And may your contribution be the first of many! :)

PR #205, opened 2021-10-31
Add QQuickPaintedItem trait and wrappers for QPainter, QPen, QBrush, QLineF

AdrianEddy:
This PR adds QQuickPaintedItem trait which allows to do custom QQuickItems painted by QPainter in pure Rust.

Wrappers around QPainter, QPen and QBrush are incomplete, but I tried to wrap most basic operations to Rust, so it's better than nothing :). If anyone needs more complex features they can always use the cpp! macro. 

BTW Is there a better way to inherit from QQuickItem in `RustObject<>` wrapper? All QQuickItem rust methods are repeated there (`classBegin` etc, in `qtdeclarative.rs:1074-1123`) but I didn't find a better way

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -1063,3 +1063,88 @@ cpp! {{
         }
     };
 }}
+
+cpp! {{
+    #include <qmetaobject_rust.hpp>
+    #include <QtQuick/QQuickItem>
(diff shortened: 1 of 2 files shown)


ogoffart:
Great change, thank you!

>  Is there a better way to inherit from QQuickItem in RustObject<> wrapper? 

Unfortunately not.
The different AbstractItemModel also have the same problem of repeated methods.

PR #206, opened 2021-11-07
Fix off-by-one error when passing `argc` to `QmlEngineHolder`'s const…

lu-fennell:
…ructor.

I'm pretty sure `argc` should be exactly the same as `arguments.len()`. E.g., in the case of passing no arguments, `arguments` contains just the command name and `argc` should be `1`. But currenty passing no arguments results in `argc == 0` (which is actually forbidden according to the docs for [QApplication](https://doc.qt.io/qt-5/qapplication.html#QApplication))

I noticed this in my app, because Qt would not set the `WM_CLASS` property when starting the app without arguments... and when passing a dummy argument (and thereby setting `argc` to `1`) fixed this problem.


ogoffart:
looks good. thank you

PR #207, opened 2021-11-07
qttypes: search for qmake-qt5 in build script

### LINKED ISSUE #197: qttypes build failure on Fedora
Fedora ships the qmake executable as /usr/bin/qmake-qt5. qttypes only searches for an executable named `qmake`.

```
error: failed to run custom build command for `qttypes v0.2.3`

Caused by:
  process didn't exit successfully: `/home/be/sw/whisperfish/target/debug/build/qttypes-fb973ac7213bde24/build-script-build` (exit status: 101)
  --- stdout
  cargo:rerun-if-env-changed=QT_INCLUDE_PATH
  cargo:rerun-if-env-changed=QT_LIBRARY_PATH

  --- stderr
  thread 'main' panicked at 'Error: Failed to execute qmake. Make sure 'qmake' is in your path!
  Os { code: 2, kind: NotFound, message: "No such file or directory" }', /home/be/.cargo/registry/src/github.com-1ecc6299db9ec823/qttypes-0.2.3/build.rs:113:21
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: build failed
```

This is easy to work around by creating a symlink called `qmake` somewhere in PATH pointing to /usr/bin/qmake-qt5.

How should this be handled by qttypes? A hack could be added to look for qmake-qt5 just for Fedora. But looking at qttype's build.rs, it doesn't actually build anything with qmake; it only uses qmake to locate Qt. So maybe it would be better to use another method to locate Qt. `find_package(Qt5)` with CMake works fine to locate Qt5 on Fedora, however that would bring in a dependency on CMake which is not necessarily present on every OS where Qt is installed.

Be-ing:
Fixes #197

### DIFF
--- a/qttypes/build.rs
+++ b/qttypes/build.rs
@@ -24,8 +24,19 @@ use std::process::Command;
 use semver::Version;
 
 fn qmake_query(var: &str) -> Result<String, std::io::Error> {
-    let qmake = std::env::var("QMAKE").unwrap_or("qmake".to_string());
-    let output = Command::new(qmake).args(&["-query", var]).output()?;
+    let output = match std::env::var("QMAKE") {
+        Ok(env_var_value) =>
+            Command::new(env_var_value).args(&["-query", var]).output(),
+        Err(_env_var_err) =>
+            Command::new("qmake").args(&["-query", var]).output().or_else(|command_err| {
+                // Some Linux distributions (Fedora, Arch) rename qmake to qmake-qt5.
+                if command_err.kind() == std::io::ErrorKind::NotFound {
+                    Command::new("qmake-qt5").args(&["-query", var]).output()
+                } else {
+                    Err(command_err)
+                }
+            }),
+    }?;
     if !output.status.success() {
         return Err(std::io::Error::new(
             std::io::ErrorKind::Other,


Be-ing:
FWIW, the qt6-qtbase-devel package in Fedora installs an executable named `qmake`.

ogoffart (changes requested):
Thanks for the patch!

Good idea to check for ErrorKind::NotFound and then try other variant.

However I don't want to panic if this is still not found. 
Maybe use `or_else` and still forward the original error.

Also we probably shouldn't do this fallback if the QMAKE env variable was set.

Be-ing:
Thanks for the suggestion to use `or_else`, that works better. I refactored it to only use the `qmake-qt5` fallback if the QMAKE environment variable is not set.

ogoffart (approved):
Thank you! looks good now.

Be-ing:
@ogoffart I see you tagged a 0.2.6 release but [crates.io](https://crates.io/crates/qmetaobject) still shows 0.2.5 as the latest version. Could you publish 0.2.6 to crates.io?

ogoffart:
only qttypes is updated, and the latest version on crates.io has this patch

Be-ing:
ah yeah, sorry the build error I encountered is actually in Whisperfish; qmetaobject built fine

PR #208, opened 2021-11-09
Simplify the qmlextensionplugins example a bit

lu-fennell:
... by using a channel to shut down the timer thread, instead of a
condvar.

I was a little distracted by the verbose handling of the condvar when looking at the example. This is my attempt to make it a bit clearer. Maybe you like it.

### DIFF
--- a/examples/qmlextensionplugins/src/lib.rs
+++ b/examples/qmlextensionplugins/src/lib.rs
@@ -1,17 +1,11 @@
 use std::ffi::CStr;
-use std::sync::{Arc, Condvar, Mutex};
-use std::thread::JoinHandle;
+use std::sync::mpsc::SyncSender;
+use std::sync::mpsc::{self, RecvTimeoutError};
 
 use chrono::Timelike;
 use cstr::cstr;
 
 use qmetaobject::prelude::*;
-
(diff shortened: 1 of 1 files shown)


lu-fennell:
Note: I used the channel from `std::sync::mpsc` which I guess is not really recommended, as it seem to have bugs. So maybe it's better to switch  to `crossbeam`'s channel.

ogoffart:
To me, the two are equal in term of complexity. Although yours has less lines of code.

I wouldn't want to pull a crossbeam dependency just for that.

Anyway, the change looks good to me.

@ratijas what's your opinion?

lu-fennell:
> To me, the two are equal in term of complexity.

Fair enough :) 

Fwiw, I, personally, find that the channel version has less subtleties to be aware of. In particular, it took me a little while to realize (resp. read up on) the interaction between condvars and mutexes and why spurious wake-ups aren't a problem here. But that's just me, I guess.

Anyhow, in case you prefer the original condvar version, it would be good to replace `thread: Option<(JoinHandle<()>, Arc<AbortCondVar>)>` with `thread: Option<Arc<AbortCondVar>>`; the `JoinHandle<()>` part is never used...

PR #209, opened 2021-11-12
Add wrappers around `QJsonValue`, `QJsonArray` and `QJsonObject`

AdrianEddy:
This PR adds basic wrappers around `QJsonValue`, `QJsonArray` and `QJsonObject` and registers the types in QMetaObject, allowing them to be used in `qt_method!`, `qt_property!` and `qt_signal!` macros

### DIFF
--- a/qmetaobject/src/qmetatype.rs
+++ b/qmetaobject/src/qmetatype.rs
@@ -488,6 +488,9 @@ impl QMetaType for QVariant {
     }
 }
 qdeclare_builtin_metatype! {QModelIndex => 42}
+qdeclare_builtin_metatype! {QJsonValue => 45}
+qdeclare_builtin_metatype! {QJsonObject => 46}
+qdeclare_builtin_metatype! {QJsonArray => 47}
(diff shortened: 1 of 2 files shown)


ogoffart:
looks good! thank you.

Could you  add tests for the added types?

AdrianEddy:
In order to add meaningful tests I had to add more complete API. My main interest was primarily to pass string hashmaps from Rust to QML in a qt signal, so at first I wrapped just the needed minimum. 

I now updated it all to wrap basic API and also added `to_json` methods. I didn't add a wrapper for `int`/`qint64` constructor and methods for `QJsonValue` because it's stored as double and can't store the full `i64` anyway, so I decided it will be better for Rust users to just have `f64` wrapper and not have any surprises trying to store `i64` range. Qt's own types are confusing too, constructor takes both `int` and `qint64`, then `toInt` returns just `int` and it's all stored as `double` anyway.

ogoffart:
Thanks for the patch!

PR #210, opened 2021-11-15
Added QColor and refactored some primitives out of lib.rs

ericcao119:
This PR greatly increases API coverage for `QColor` and also refactors out some primitives in `lib.rs`.


ogoffart:
Thanks for the PR.
Moving the classes in different module is a good change.
However that makes it difficult to see exactly what changed in the code. I'd appreciate if the commit were split to one commit that just do moving but doesn't change any code, and another commit that add or change the code
In particular, i just see from the tests changes that this change the API of from_rgb_f which is a breaking change. I'd prefer avoiding breaking change if possible.

PR #211, opened 2021-11-15
Allow using `QGuiApplication` and not linking to `QtWidgets`

AdrianEddy:
This PR adds a new feature flag that allows to skip linking the QtWidgets module and use `QGuiApplication` instead of `QApplication`. I added it as an optional feature that disables the linking to not break any existing code and feature set that users may already have.

### DIFF
--- a/qmetaobject/Cargo.toml
+++ b/qmetaobject/Cargo.toml
@@ -15,6 +15,7 @@ repository = "https://github.com/woboq/qmetaobject-rs"
 default = ["log"]
 chrono_qdatetime = ["qttypes/chrono"]
 webengine = ["qttypes/qtwebengine"]
+nowidgets = ["qttypes/nowidgets"]
 
 [dependencies]
 qttypes = { path = "../qttypes", version = "0.2.4", features = ["qtquick"] }
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -24,11 +24,21 @@ use crate::*;
 /// So this is a guard that will be used to panic if the engine is created twice
 static HAS_ENGINE: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
 
+#[cfg(feature = "nowidgets")]
+cpp! {{
+    #include <QtGui/QGuiApplication>
+    #define QAPPLICATION QGuiApplication
+}}
+#[cfg(not(feature = "nowidgets"))]
+cpp! {{
+    #include <QtWidgets/QApplication>
+    #define QAPPLICATION QApplication
+}}
+
 cpp! {{
     #include <memory>
     #include <QtQuick/QtQuick>
     #include <QtCore/QDebug>
-    #include <QtWidgets/QApplication>
     #include <QtQml/QQmlComponent>
 
     struct SingleApplicationGuard {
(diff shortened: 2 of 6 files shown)


qttypes/Cargo.toml
```diff
@@ -31,6 +31,8 @@ qtmultimediawidgets = []
 qtsql = []
 # Link against QtTest
 qttest = []
+# Don't link QtWidgets and use QGuiApplication
+nowidgets = []
```
ogoffart, line 35: In Rust, features need to be additive.
So this should be a `qtwidget` feature and enable it by default.

AdrianEddy: But then how about users who already have no-default-features enabled? It would break the build for them

ogoffart: Right.
I suppose we will have to increase the semver then.

ogoffart: Something like https://github.com/rust-lang/rfcs/pull/3146  would be nice.

AdrianEddy:
Hmm, looks like the `#[cfg(feature)]` macros don't apply to `cpp!` macro, because both cases seem to be evaluated and `QAPPLICATION` is always defined as the latter one, regardless of the selected features. Any idea how this could be resolved?

qmetaobject/src/qtdeclarative.rs
```diff
@@ -24,11 +24,21 @@ use crate::*;
...
+cpp! {{
+    #include <QtWidgets/QApplication>
+    #define QAPPLICATION QApplication
+}}
+#[cfg(not(feature = "widgets"))]
+cpp! {{
+    #include <QtGui/QGuiApplication>
+    #define QAPPLICATION QGuiApplication
+}}
+
```
AdrianEddy, line 37: Hmm, looks like the #[cfg(feature)] macros don't apply to cpp! macro, because both cases seem to be evaluated and QAPPLICATION is always defined as the latter one, regardless of the selected features. Any idea how this could be resolved?

AdrianEddy, line 37: Hmm, looks like the #[cfg(feature)] macros don't apply to cpp! macro, because both cases seem to be evaluated and QAPPLICATION is always defined as the latter one, regardless of the selected features. Any idea how this could be resolved?

ogoffart: maybe add a `-DNO_WIDGETS` from the build.rs so `#ifdef` can be used there?

AdrianEddy: That works, thank you! 👌

ogoffart (commented):
Thanks, the change starts to look good.

Since this require a semver version change. I'm wondering about merging this now or waiting a bit to see if there are other needed breaking changes.  How much do you require this change?

AdrianEddy:
> Since this require a semver version change. I'm wondering about merging this now or waiting a bit to see if there are other needed breaking changes. How much do you require this change?

It's not super important, it'll be just nice to save some space by not having to deploy QtWidgets module when it's not used. It can wait, it's not a blocker for me

qmetaobject/Cargo.toml
```diff
@@ -12,9 +12,10 @@ keywords = ["Qt", "QML", "QMetaObject",]
...
 
 [features]
-default = ["log"]
+default = ["log", "widgets"]
 chrono_qdatetime = ["qttypes/chrono"]
 webengine = ["qttypes/qtwebengine"]
+widgets = ["qttypes/qtwidgets"]
 
 [dependencies]
 qttypes = { path = "../qttypes", version = "0.2.5", features = ["qtquick"] }
```
lights0123, line 21:
```suggestion
qttypes = { path = "../qttypes", version = "0.2.5", default-features = false, features = ["required", "qtquick"] }
```

This PR is extremely helpful for me: I'm creating an app for the reMarkable e-ink tablet, which only supports QtQuick. However, `qttypes` still tries to link to QtWidgets with its default features, which causes an error for me because it doesn't exist.

ogoffart:
Maybe it's soon time to make a breaking change release so we can merge this.

PR #212, opened 2021-11-16
Refactored qreal, QByteArray, QString, and QUrl into different files

ericcao119:
Refactored out some of the core types into their own files


ogoffart:
Tahnks!

PR #213, opened 2021-11-16
Refactored out QColor into its own module and increased Qcolor API coverage

ericcao119:
This version does not break the current API. My only reservation is about how to handle default arguments going forward. This class has a lot of redundant methods like rgb_f, rgba_f, cymk_f, cymka_f that exist because of rust's lack of default arguments. Is the preference to use an Option or to just create multiple methods in the future.

### DIFF
--- a/qttypes/src/gui/qcolor.rs
+++ b/qttypes/src/gui/qcolor.rs
@@ -557,7 +557,7 @@ impl QColor {
         })
     }
 
-    pub fn name_format(&self, format: NameFormat) -> QString {
+    pub fn name_with_format(&self, format: NameFormat) -> QString {
         cpp!(unsafe [self as "const QColor*", format as "QColor::NameFormat"] -> QString as "QString" {
             return self->name(format);
         })
--- a/qttypes/src/gui/qcolor.rs
+++ b/qttypes/src/gui/qcolor.rs
@@ -1,0 +1,871 @@
+pub use crate::core::{qreal, QString};
+
+#[cfg(not(no_qt))]
+use cpp::{cpp, cpp_class};
+
+cpp! {{
+    #include <QtGui/QColor>
+    #include <QtCore/QString>
+}}
+
+#[derive(Copy, Clone, Eq, PartialEq, Hash)]
+pub struct QRgb(u32);
+
+impl QRgb {
+    pub fn alpha(&self) -> u8 {
+        ((self.0 >> 24) & 0x000000ff) as u8
+    }
+    pub fn red(&self) -> u8 {
+        ((self.0 >> 16) & 0x000000ff) as u8
+    }
+    pub fn green(&self) -> u8 {
+        ((self.0 >> 8) & 0x000000ff) as u8
+    }
+    pub fn blue(&self) -> u8 {
+        (self.0 & 0x000000ff) as u8
+    }
+}
+
+impl From<u32> for QRgb {
+    fn from(val: u32) -> QRgb {
+        QRgb(val)
+    }
+}
+
+impl Into<u32> for QRgb {
+    fn into(self) -> u32 {
+        self.0
+    }
+}
+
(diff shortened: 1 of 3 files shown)


ogoffart:
Looks good.
Can you rebase?

qttypes/src/gui/qcolor.rs
```diff
@@ -0,0 +1,871 @@
...
+        })
+    }
+
+    pub fn name(&self) -> QString {
+        cpp!(unsafe [self as "const QColor*"] -> QString as "QString" {
+            return self->name();
+        })
+    }
+
+    pub fn name_format(&self, format: NameFormat) -> QString {
```
ogoffart, line 560: maybe this should be renamed `name_with_format`  or just have `name` that always take a format.
another option would be to just have a `name_argb` and `name_rgb`.

qttypes/src/gui/qcolor.rs
```diff
@@ -0,0 +1,871 @@
...
+    HexArgb = 1,
+}
+
+/// Bindings for [`QColor::Spec`][class] enum class.
+///
+/// [class]: https://doc.qt.io/qt-5/qcolor.html#Spec-enum
+#[repr(u32)]
+#[derive(Clone, Copy, PartialEq, Debug)]
+#[allow(non_camel_case_types)]
+pub enum Spec {
```
ogoffart, line 89: This should be named QColorSpec,  and NameFormat  QColorNameFormat, because it otherwise is not descriptive enough as it is outside of a namespace.

qttypes/src/gui/qcolor.rs
```diff
@@ -0,0 +1,871 @@
...
+}
+
+/// Bindings for [`QColor::NameFormat`][class] enum class.
+///
+/// [class]: https://doc.qt.io/qt-5/qcolor.html#NameFormat-enum
+#[repr(u32)]
+#[derive(Clone, Copy, PartialEq, Debug)]
+#[allow(non_camel_case_types)]
+pub enum NameFormat {
+    // #RRGGBB A "#" character followed by three two-digit hexadecimal numbers (i.e. #RRGGBB).
```
ogoffart, line 77: should be doc comment. (`///`)

ogoffart (changes requested):
I still have a concern about the naming of Spec and NameFormat since they are re-exported at the top level, and their raw name is lacking context.


Regarding name, i'm a bit unsure if this is good api, but since that's also what Qt has, let's just go with that. (In my mind, this crate shouldn't be a 1:1 binding of Qt, but offer idiomatic rust API for using QML. So one does not have to just blindly take Qt API, but think about the best and more efficient naming.  But in that case i think it's ok to just follow Qt)

ericcao119:
I'm fine with the name change but I do think that name_with_format and name should be kept as separate functions since name_with_format allows you to specify the output format.

By idiomatic, do you mean using rust types for the most part? Like returning String rather than QString. Because otherwise I think the API is pretty straightforward.

ericcao119:
Also, NameFormat and Spec don't need to be re-exported at the top level if that is what you prefer.

ogoffart:
> By idiomatic, do you mean using rust types for the most part? Like returning String rather than QString. Because otherwise I think the API is pretty straightforward.

I mean it is a tradeof between
 1. performance
 2. idiomatic
 3. mapping of the Qt API

And I try to keep that order.  So we use QString instead of String to avoid needless conversions. But for naming and what we expose, we prefer rust naming and convention to Qt/C++ ones.

ogoffart:
I pushed by mistake your commits into the master branch in e188dea9486c756970ef91d328eefd5b8f153f0b
as i had them locally to do some tests.  
So i'll close this PR (since it is merged, in fact) I'll do the last adjustments myself.

Thanks for the contribution.

PR #214, opened 2021-11-21
Updated Q

ericcao119:
QRgba64 is now derived from the C++ class rather than implemented in Rust.
Additionally, some conversions were added.
Conversion from u64 to QRgba64 is not currently implemented due to endian-ness concerns.

### DIFF
--- a/qttypes/src/gui/qcolor.rs
+++ b/qttypes/src/gui/qcolor.rs
@@ -6,7 +6,7 @@ cpp! {{
     #include <QtCore/QString>
 }}
 
-#[derive(Copy, Clone, Eq, PartialEq, Hash)]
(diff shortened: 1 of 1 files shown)


ogoffart:
> QRgba64 is now derived from the C++ class rather than implemented in Rust.

What's the rationale?  Having it implemented in rust if probably much more efficient for such low level API.  When using QRgba64, one want to use it as light class, and it ideally should be Copy and use bit shuffling to extract the color without going through a function call in C++

ericcao119:
If we do that you might have to change the license to LGPL.

We could create a Rust version but it would need to be compatible and have the same internal representation since it is used for SIMD applications, which makes many of the operations exactly identical in code. I could do it if you really want but it seems a bit strange to do it this way when the implementation is already there.

PR #218, opened 2021-12-15
Added Methods to QStringList.

Ayush1325:
Added methods for constructing QStringList from Rust Side.
Some methods in [ki18n-rs](https://github.com/Ayush1325/ki18n-rs) have QStringList as arguments and thus there need to be methods that allow the construction from Rust Side.

### DIFF
--- a/qttypes/src/lib.rs
+++ b/qttypes/src/lib.rs
@@ -1661,7 +1661,7 @@ impl QStringList {
         });
     }
 
-    pub fn append(&mut self, value: QString) {
+    pub fn push(&mut self, value: QString) {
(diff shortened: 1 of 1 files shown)


ogoffart:
looks good.
I think we should use the same API as Rust.  So `push` instead of `append` and `remove` instead of `remove_at`,  like `Vec` does.

ogoffart:
thank you!
One last thing.  Could you write a small `#[test]` that uses these functions

ogoffart (approved):

ogoffart:
thanks!

PR #220, opened 2022-01-02
Add wrapper around qmlRegisterModule function

### LINKED ISSUE #219: Missing API wrapper for qmlRegisterModule
## Summary 

There is no Rust function for [qmlRegisterModule](https://doc.qt.io/qt-5/qqmlengine.html#qmlRegisterModule) Qt method.

ratijas:
This is a rarely used feature, but is still needed in some corner-cases.

Closes #219

### DIFF
--- a/qmetaobject/src/lib.rs
+++ b/qmetaobject/src/lib.rs
@@ -216,6 +216,8 @@ pub mod webengine;
 pub mod prelude {
     #[cfg(qt_5_8)]
     pub use crate::qtdeclarative::qml_register_enum;
+    #[cfg(qt_5_9)]
+    pub use crate::qtdeclarative::qml_register_module;
     pub use crate::{
         qml_register_type, qrc, qt_base_class, qt_method, qt_plugin, qt_property, qt_signal,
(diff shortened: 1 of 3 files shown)


ratijas:
The test should also be guarded with cfg.  Gotta fix tomorrow.

ratijas:
Just added little clean-ups while I'm at it. Should be good to go now. @ogoffart ?

ogoffart:
I don't know... I'm not convinced this is so useful...  but why not?

ratijas:
What puzzles me, is that Qt6 didn't drop version major/minor arguments nor provided a version-less overload. Actually, this is true for all registrators listed in https://doc.qt.io/qt-6/qqmlengine.html#qmlRegisterModule — feels like an incomplete refactoring or just a backward compatibility, I don't know…

ratijas:
GitHub is weird. I used "rebase & merge" strategy, because it wouldn't make to do a branch-merge commit here. This PR was already on the tip of master, but GitHub _still insisted_ on rebasing and re-committing, so now commits have slightly different author&committer fields and lost signatures. What the heck?…

Issue #221, opened 2022-01-03
A couple of QJSValue methods is missing


gikari:
## Summary

QJSValue is missing the following methods:
- [ ] [call methods](https://doc.qt.io/qt-5/qjsvalue.html#call) in case the QJSValue is a JS function (call, callAsConstructor, callWithInstance, isCallable).
- [ ] [property methods](https://doc.qt.io/qt-5/qjsvalue.html#property) (property by name/by index, set by name/by index, has, hasOwn, delete)
- [ ] [is methods](https://doc.qt.io/qt-5/qjsvalue.html#isBool) for various types (bool, Number, etc)

PR #222, opened 2022-01-05
feat(qmetaobject): add basic "is" wrappers for QJSValue

gikari:
I wasn't able to add wrapper for other types, since I couldn't construct values for unit-tests using From trait.

Ref #221


ogoffart:
Thanks!

PR #224, opened 2022-01-06
feat(qmetaobject): add "is" wrappers for special QJSValues

gikari:
I hope, I handled `enum` right.

Ref #221


qmetaobject/src/qtdeclarative.rs
```diff
@@ -506,11 +506,7 @@ pub fn qml_register_type<T: QObject + Default + Sized>(
 ///
 /// [qt]: https://doc.qt.io/qt-5/qqmlengine.html#qmlRegisterModule
 #[cfg(qt_5_9)]
-pub fn qml_register_module(
-    uri: &CStr,
-    version_major: u32,
-    version_minor: u32,
-) {
+pub fn qml_register_module(uri: &CStr, version_major: u32, version_minor: u32) {
```
ratijas, line 509: huh?  autoformatter?

ratijas:
LGTM, except unrelated `qml_register_module` formatting change.

qmetaobject/src/qtdeclarative.rs
```diff
@@ -926,13 +922,29 @@ cpp_class!(
     pub unsafe struct QJSValue as "QJSValue"
 );
 
+/// Wrapper for [`QJSValue::SpecialValue`][qt]
+///
+/// [qt]: https://doc.qt.io/qt-5/qjsvalue.html#SpecialValue-enum
+#[repr(u32)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum QJSValueSpecialValue {
```
ogoffart, line 930: I don't know if it makes sense to expose this in rust.

I would prefer having functiopn like 

`QJSValue::null()`  to create a null value, and `QJSValue::default()` creates an undefined value  (we could also add `undefined()` 

What do you think?

ogoffart: But public and documented in Qt doesn't mean there is  reason to expose them in this crate.

ratijas: :thinking: konsistency? After all, what are the downsides?

I have to admit, I'm not a big fan of a lengthy and repetitive name `QJSValueSpecialValue`, but that's due to a technical restriction in Rust. Maybe `QJSSpecialValue` would be marginally better, since it's not 1:1 mapping anyway?

qmetaobject/src/qtdeclarative.rs
gikari: Yes, I separated it into another commit, but I can remove it, if necessary.

qmetaobject/src/qtdeclarative.rs
gikari: I agree, that `null` and `undefined` could be useful.

However, I also think having consistency with Qt Framework is important. If someone decides to use the crate, they would expect a 1:1 mapping of entities, that come from Qt. They may also find the enum useful (for some reason).

gikari:
It appears, that `Default` is already [used](https://docs.rs/cpp/0.5.6/cpp/macro.cpp_class.html#derived-traits) by cpp_class! macros. I do not know if it's fixable.

ratijas:
> It appears, that `Default` is already [used](https://docs.rs/cpp/0.5.6/cpp/macro.cpp_class.html#derived-traits) by cpp_class! macros. I do not know if it's fixable.

From the link you referenced:

>> and `Default` if the class is default constructible

QJSValue does not explicitly define a standalone default constructor, but it does have one with a single optional argument:

```c++
QJSValue::QJSValue(QJSValue::SpecialValue value = UndefinedValue)
```

so a `QJSValue::default()` constructor should already be the same as calling `undefined()`. Worth a test though :)

gikari:
> so a `QJSValue::default()` constructor should already be the same as calling `undefined()`. Worth a test though :)

Added as an `assert` to is_undefined test.

gikari:
Ping?

PR #226, opened 2022-01-16
Added QStandardPaths::StandardLocation enum.

Ayush1325:
The documentation about this enum can be found [here](https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum). I need this enum for [konfig](https://invent.kde.org/oreki/kconfig-rs). 
I also have a few questions about how to define this enum:
1. Naming: What is the naming convention for enums in qmetaobject. `QStandardPathsStandardLocation` seems too long which was followed for `QColorSpec` and a few other enums.
2. Multiple fields with the same value: C++ allows multiple enum fields to have the same integer value which is not allowed in Rust. In this enum, `AppLocalDataLocation` has the same value as `DataLocation`. Currently, I have defined it using a const but not sure if that is the correct way to do this.

### DIFF
--- a/qttypes/src/lib.rs
+++ b/qttypes/src/lib.rs
@@ -1222,6 +1222,34 @@ impl QPen {
     //    qreal	widthF() const
 }
 
+/// Bindings for [`QStandardPaths::StandardLocation`][enum] enum.
+///
+/// [enum]: https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum
+#[repr(C)]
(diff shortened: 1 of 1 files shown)


ogoffart:
QStandardPaths::DataLocation is deprecated, so just don't have that, and keep only AppLocalDataLocation

ogoffart:
> QStandardPathsStandardLocation seems too long

Yeah, maybe `QStandardPathLocation` ?

ogoffart:
Thank you

PR #230, opened 2022-01-20
Implemented From for QStringList.

Ayush1325:
Can now do:
```rust
QStringList::from(["Three", "Two"]);
QStringList::from(["Three".to_string(), "Two".to_string()]);
QStringList::from([QString::from("Three"), QString::from("Two")]);
```

### DIFF
--- a/qttypes/src/lib.rs
+++ b/qttypes/src/lib.rs
@@ -1760,6 +1760,20 @@ where
     }
 }
 
+impl<T> From<&[T]> for QStringList
+where
+    T: Clone,
+    QString: From<T>,
+{
+    fn from(s: &[T]) -> Self {
+        let mut list = QStringList::new();
+        for i in s {
+            list.push(QString::from(i.clone())); // i: &T in case of slice.
+        }
+        list
+    }
+}
+
 impl<T> From<QStringList> for Vec<T>
 where
     T: Clone,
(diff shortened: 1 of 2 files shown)


Ayush1325:
Const Generics were added in rustc 1.57.0, I think. Currently, 1.58.0 has also been released. So do I need to find a workaround that works on 1.56.0 and older or not?
The standard library `Vec::from()` and `Hashmap::From` also seem to use const generics now.

qttypes/src/lib.rs
```diff
@@ -1734,6 +1734,46 @@ impl fmt::Debug for QStringList {
...
+    fn from(s: [T; N]) -> Self {
+        let mut list = QStringList::new();
+        for i in s {
+            list.push(QString::from(i));
+        }
+        list
+    }
+}
+
+impl<T> From<Vec<T>> for QStringList
```
ogoffart, line 1750: should it maybe be `From<&[T]>`  ?

ogoffart:
> So do I need to find a workaround that works on 1.56.0 and older or not?

I think it's fine to upgrade the MSRV.
Then we can use the rust-version field in the Cargo.toml

qttypes/src/lib.rs
Ayush1325: Well, it can be done but since `QString`, `String` do not implement Copy, the new impl will look like this:
```rust
impl<T> From<&[T]> for QStringList
where
    T: Clone,
    QString: From<T>,
{
    fn from(s: &[T]) -> Self {
        let mut list = QStringList::new();
        for i in s {
            list.push(QString::from(i.clone()));
        }
        list
    }
}
```
There might be a way to implement this without clone, but I'm not sure how to do it. The vector implementation alone does not use any clone. So I can add the slice implementation too, but I don't think it should replace the vector implementation, at least not before we can get rid of all the cloning.

ogoffart: Cloning a QString is cheap (just increment a reference count)
But creating a Vec is not cheap (allocation)
If you have a `Vec<T>` you can get a `&[T]` for free. But if you have a `&[T]`  you need to allocate to create a Vec.

Ayush1325: Hmm. So should I remove the Vec implementation or just add the slice implementation leaving the Vec one intact?

ogoffart (approved):
Looks good to me.
I don't know why appveyor is still using an old rust version... it should use stable. Maybe there is something wrong with the cache?
I wonder to disable the appveyor then

ogoffart:
I remved the appveyor build so we can merge this.

PR #231, opened 2022-01-21
Changed return type to Self::Output in Index trait

Ayush1325:
Shouldn't really change any functionality.


ogoffart:
Why does it matter?

Anyway, since it can't hurt, i'm just merging this.

Ayush1325:
> Why does it matter?
> 
> Anyway, since it can't hurt, i'm just merging this.

Just convention, I guess. I think it is better to follow convention unless there is a reason not to.

Issue #232, opened 2022-01-22
[Feature Request] Wrapper for QMap<K, T>


Ayush1325:
# Introduction
I am currently working on a wrapper for the KConfig KDE framework and need to return `QMap<K, T>` from a function. It seems there are no implementations of QT generic types in qmetaobject yet. I will try to tackle this myself at some point but it seems like it will take a lot more time than I currently have. So just opening this issue if anyone else wants to give it a try.

# Helpful Links
- [CxxVector\<T\>](https://docs.rs/cxx/latest/cxx/struct.CxxVector.html): Binding to C++ std::vector<T, std::allocator<T>>.

ogoffart:
We can't have template code like that. So you would need to wrap some specific one, (for example `QVariantMap` or `QMapFooBar as "QMap<Foo, Bar>"` for a specific Foo and Bar)
If QMap would have type erasure, we could do it like CxxVector, but it doesn't so we can't have it.

Anyway, i believe, for your usecase the best is not to make a 1-1 API with the C++ API, but return something that make sense for rust. 

 So i guess `QVariantMap` would kind of work, would it not?

Ayush1325:
> 

`QVariantMap` might work. But I will need the ability to convert to and from Hashmap in Rust. 
I did not know it was not possible to have something like generics in Qt types. I guess I will have to study more about tye erasures,

PR #234, opened 2022-01-28
Added QVariantMap Type

Ayush1325:
I'm not quite sure how to implement Iterator for QVariantMap. Any ideas?

Also, can someone check my implementation of `Index` trait. The `find()` method in QMap seems to return an iterator even though `insert()` replaces the old value if the key exists. So how can there be multiple values in the same key?

Finally, I'm not quite sure how to implement Iterator for QVariantMap.

Related to #232

### DIFF
--- a/qttypes/src/lib.rs
+++ b/qttypes/src/lib.rs
@@ -122,7 +122,7 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 use std::collections::HashMap;
 use std::convert::From;
-use std::fmt;
+use std::fmt::{self, Write};
 use std::iter::FromIterator;
 use std::ops::{Index, IndexMut};
 
@@ -557,6 +557,13 @@ impl QVariant {
 
     // FIXME: do more wrappers
 }
+
+impl fmt::Debug for QVariant {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str(self.to_qbytearray().to_string().as_str())
+    }
+}
+
 impl From<QString> for QVariant {
     /// Wrapper around [`QVariant(const QString &)`][ctor] constructor.
     ///
@@ -881,6 +888,7 @@ cpp_class!(
     /// Wrapper around [`QVariantMap`][type] typedef.
     ///
     /// [type]: https://doc.qt.io/qt-5/qvariant.html#QVariantMap-typedef
+    #[derive(Default, PartialEq, Eq)]
     pub unsafe struct QVariantMap as "QVariantMap"
 );
 
@@ -970,6 +978,12 @@ impl IndexMut<QString> for QVariantMap {
                 return &(*self)[key];
             })
         }
+    }
+}
+
+impl fmt::Debug for QVariantMap {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_map().entries(self.into_iter()).finish()
     }
 }
 
@@ -1108,7 +1122,6 @@ mod qvariantmap_tests {
 
         for (k, v) in map.into_iter() {
             assert_eq!(hashmap[k.to_string().as_str()].to_string(), v.to_qbytearray().to_string());
-            // println!("Key: {}, Value: {}", k, v.to_qbytearray().to_string());
         }
     }
 }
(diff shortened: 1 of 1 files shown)


qttypes/src/lib.rs
```diff
@@ -875,6 +875,118 @@ mod tests {
...
+impl Index<QString> for QVariantMap {
+    type Output = QVariant;
+
+    /// Wrapper around [`at(int)`][method] method.
+    ///
+    /// [method]: https://doc.qt.io/qt-5/qlist.html#at
+    fn index(&self, key: QString) -> &QVariant {
+        assert!(self.contains(key.clone()));
+        unsafe {
+            &*cpp!([self as "QVariantMap*", key as "QString"] -> *const QVariant as "const QVariant*" {
```
ogoffart, line 950: I'd return `-> Option<&QVariant> as "const QVariant*"`  
and then do a `.expect("key not in the QVariant")`  instead of the assert.  that way you only do the lookup once.

You might also need a `#[track_caller]` so the panic location appears for the user who uses that function

ogoffart (commented):
Looks good so far

ogoffart:
Regarding the Iterator, maybe you can have a

```rust
cpp_class!(unsafe struct QVariantMapIteratorInternal as "QVariantMap::iterator");
pub struct QVariantMapIterator<'a> {
     map: &'QVariantMap,
     iterator: QVariantMapIteratorInternal,
}
```

I did not try that.

qttypes/src/lib.rs
Ayush1325: Using `.find().value()` seems to return a non-null garbage pointer.

ogoffart:
```suggestion
            &*cpp!([self as "QVariantMap*", key as "QString"] -> *const QVariant as "const QVariant*" {
```

Actually, you must use `self as "const QVariantMap*"`   or `constFind`

ogoffart: (and then compare the result with constEnd

ratijas:
Nit: consider imperative tone rather than past tense in PR and commit title, e.g.: "**Add** QVariantMap Type". Upstream Qt have some nice guidelines about it too :)

Ayush1325:
I think I have added everything I needed now. The appveyor build fails for the same reason as #230.

qttypes/src/lib.rs
```diff
@@ -875,6 +883,336 @@ mod tests {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_map().entries(self.into_iter()).finish()
+    }
+}
+
+cpp_class!(unsafe struct QVariantMapIteratorInternal as "QVariantMap::iterator");
+
+/// Internal class used to iterate over a [`QVariantMap`][]
+///
+/// [`QVariantMap`]: ./struct.QVariantMap.html
```
ogoffart, line 1020:
```suggestion
/// Internal class used to iterate over a [`QVariantMap`]
```

We can use the intra-doc links.

qttypes/src/lib.rs
```diff
@@ -875,6 +883,336 @@ mod tests {
...
+pub struct QVariantMapIterator<'a> {
+    map: &'a QVariantMap,
+    iterator: QVariantMapIteratorInternal,
+}
+
+impl<'a> QVariantMapIterator<'a> {
+    fn key(&self) -> Option<&'a QString> {
+        let iterator = &self.iterator;
+        unsafe {
+            cpp!([iterator as "QVariantMap::iterator*"] -> *mut QString as "const QString*" {
```
ogoffart, line 1030:
```suggestion
            cpp!([iterator as "const QVariantMap::iterator*"] -> &QString as "const QString*" {
```
Not all the code respect that, but it would be better to respect the constness of things.
Like when we have a reference in rust, then use a const pointer in C++. and only use a non-const pointer in C++ if we have a &mut reference in rust.

Also you can just return rust references instead of pointers if you know it's not null or dangling.

ogoffart (approved):
Looks good.

With some minor comments

qttypes/src/lib.rs
Ayush1325: I cannot convert it to a reference since I am using `as_ref()` method to convert from pointer to `Option<&QString>`. I did convert it to *const QString though.

ogoffart: ah yes, good point.  You could just use `-> Option<&QString> as "const QString*"`  because Option ensure that it use the niche.

PR #236, opened 2022-01-30
Fix building on Android

AdrianEddy:
For some reason `-std=c++17` is not detected as supported flag, but it's required on Android so add it explicitly.

Please squash this when merging, sorry for the commit mess


ogoffart:
Thanks for the patch.
Do you have any idea why `flag_if_supported`  is not working properly?
A bug in the `cc` crate?

AdrianEddy:
I wasn't able to find out the reason, but I didn't dig too deep into this, it could be a bug in the `cc` crate.

ogoffart:
C++17 is enabled unconditionally now  with the 0.2.7 release

Issue #235, opened 2022-01-29
Refactor qttypes


Ayush1325:
Currently, libs.rs in qttypes basically contains most of the types. I think the types should be moved to separate modules, like QString is inside core.

Current Refactor Propsal:

- [ ] `QDate` -> `core::qdate::Qdate`
- [ ] `QTime` -> `core::qtime::QTime`
- [ ] `QDateTime` -> `core::qdatetime::QDateTime`
- [ ] `QVariant` -> `core::qvariant::QVariant`
- [ ] `QVariantList` -> `core::qlist::qvariantlist::QVariantList`
- [ ] `QModelIndex` -> `core::qmodelindex::QModelIndex`
- [ ] `QRectF` -> `core::qrectf::QRectF`
- [ ] `QPointF` -> `core::qpointf::QPointF`
- [ ] `QPoint` -> `core::qpoint::QPoint`
- [ ] `QSizeF` -> `core::qsizef::QSizeF`
- [ ] `QSize` -> `core::qsize::QSize`
- [ ] `QMargins` -> `core::qmargins::QMargins`
- [ ] `ImageFormat` -> `core::qimage::ImageFormat`
- [ ] `QImage` -> `gui::qimage::QImage`
- [ ] `QPixmap` -> `gui::qpixmap::QPixmap`
- [ ] `PenStyle` -> `core::qt::PenStyle`
- [ ] `QPen` -> `gui::qpen::QPen`
- [ ] `QStandardPathLocation` -> `core::qstandardpathlocation::QStandardPathLocation`
- [ ] `BrushStyle` -> `core::qt::BrushStyle`
- [ ] `QBrush` -> `gui::qbrush::QBrush`
- [ ] `QLineF` -> `core::qlinef::QLineF`
- [ ] `QPainter` -> `gui::qpainter::QPainter`
- [ ] `QPainterRenderHint` -> `gui::qpainter::QPainterRenderHint`
- [ ] `QJsonValue` -> `core::qjsonvalue::QJsonValue`
- [ ] `QStringList` -> `core::qlist::qstringlist::QStringList`
- [ ] `QJsonObject` -> `core::qjsonobject::QJsonObject`
- [ ] `QJsonArray` -> `core::qjsonarray::QJsonArray`

ogoffart:
makes sense.

ratijas:
[...]

PR #247, opened 2022-02-13
Add generic iterator for QList

Ayush1325:
Now both QStringList and QVariantList use a generic iterator for iteration implementation. The QListIterator is just a generic form of the `QVariantListIterator` that was already present.

This commit also refactors `QStringList` and `QVariantList` to separate modules inside qlist as mentioned in #235.


Ayush1325:
I would like someone else to comment if anything can be improved here. I think this iterator implementation should work with any wrapper over `QList<T, K>`.

ogoffart:
I guess it makes sense.

Ayush1325:
@ogoffart What is the preferred way to merge. Should I use `Rebase and merge` or `Create a merge commit`?

ogoffart:
I prefer rebase, to keep the history somewhat linear.

Issue #249, opened 2022-02-21
Failed to compile qttypes with msvc2019 and Qt 6.3.0-beta2


fluentpwn:
Here is the error:
```
 running: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Tools\\MSVC\\14.28.29910\\bin\\HostX64\\x64\\cl.exe" "-nologo" "-MD" "-Z7" "-Brepro" "-I" "C:\\Users\\\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\qttypes-0.2.6" "-I" "C:/Qt/6.3.0/msvc2019_64/include" "-W4" "/std:c++17" "/Zc:__cplusplus" "-FoC:\\Users\\\\Desktop\\projects\\cutie\\target\\debug\\build\\qttypes-7f5266fa31b11550\\out\\rust_cpp\\cpp_closures.o" "-c" "C:\\Users\\\\Desktop\\projects\\cutie\\target\\debug\\build\\qttypes-7f5266fa31b11550\\out\\rust_cpp\\cpp_closures.cpp"
  cpp_closures.cpp
  C:\Qt\6.3.0\msvc2019_64\include\QtCore/qglobal.h(105): error C2338: On MSVC you must pass the /permissive- option to the compiler.
  exit code: 2

  --- stderr


  error occurred: ToolExecError: Command "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Tools\\MSVC\\14.28.29910\\bin\\HostX64\\x64\\cl.exe" "-nologo" "-MD" "-Z7" "-Brepro" "-I" "C:\\Users\\\\.cargo\\registry\\src\\github.com-1ecc6299db9ec823\\qttypes-0.2.6" "-I" "C:/Qt/6.3.0/msvc2019_64/include" "-W4" "/std:c++17" "/Zc:__cplusplus" "-FoC:\\Users\\\\Desktop\\projects\\cutie\\target\\debug\\build\\qttypes-7f5266fa31b11550\\out\\rust_cpp\\cpp_closures.o" "-c" "C:\\Users\\\\Desktop\\projects\\cutie\\target\\debug\\build\\qttypes-7f5266fa31b11550\\out\\rust_cpp\\cpp_closures.cpp" with args "cl.exe" did not execute successfully (status code exit code: 2).
```

ogoffart:
[...]

PR #250, opened 2022-02-21
Add flag "/permissive-" for MSVC builds

fluentpwn:
Add flag "/permissive-" for MSVC builds for compatibility with Qt 6.3.0, solves issue #249


ogoffart:
Thanks

ogoffart:
Actually, I believe this might now be enough because the dependent crates will also need this flag .

There starts to be quite a lot of flags, qttype should use a metadata to tell dependent crate which flags to use

PR #255, opened 2022-02-26
Implement Add and AddAssign for QString

Ayush1325:
Uses the `QString &QString::append(const QString &str)` internally.


Ayush1325:
`QString &QString::append(const QString &str)` appends to the current QString as well as returns a new QString. So is it fine the way it is right now or should I change the method to something like, `void QString::append(const QString &str)`?

PR #256, opened 2022-03-09
Add Qt::GlobalColor enum wrapper.

Ayush1325:
Also covers #235


ogoffart:
The patch looks good.
That said, i'm wondering what's the use case for this in rust.
In C++ it makes sense because one can just write `Qt::Red` where one expect a color, and we have implicit conversion.

In rust, a, ideal api would be something like an associated const (`QColor::RED`)  but that's not possible to have QColor as const. 
I was also thinking an associated function like `QColor::red()` but that's conflicting with the existing accessor.

So I guess what you have done is indeed the best if we really want to create colors like this.

Ayush1325:
> The patch looks good. That said, i'm wondering what's the use case for this in rust. In C++ it makes sense because one can just write `Qt::Red` where one expect a color, and we have implicit conversion.
> 
> In rust, a, ideal api would be something like an associated const (`QColor::RED`) but that's not possible to have QColor as const. I was also thinking an associated function like `QColor::red()` but that's conflicting with the existing accessor.
> 
> So I guess what you have done is indeed the best if we really want to create colors like this.

Well, I encountered this in the [Introduction to KConfig](https://develop.kde.org/docs/use/configuration/introduction/), which I am rewriting for Rust. If I say the use case, then I prefer using an enum compared to using a string color name. The GlobalColors enum seems to contain the common colors that are needed in the applications, so I think having something similar would be great.

ogoffart (approved):
ok, makes sense

PR #261, opened 2022-03-13
Impl From<QString>, From<QByteArray> for Vec<u8>

Ayush1325:
I basically needed implementation of `Into<Vec<u8>>` so that I can construct `CString::new()` from `QString`. 

I am not sure whether this is the best way to achieve this so I would like someone else's input on this.

### DIFF
--- a/qttypes/src/qtcore/qstring.rs
+++ b/qttypes/src/qtcore/qstring.rs
@@ -407,4 +407,13 @@ mod tests {
         let str3 = QString::from("abcef") + QString::from("gefg");
         assert_eq!(str1, str3);
     }
+
+    #[test]
+    fn cstring() {
+        use std::ffi::CString;
+
+        let s = "abc";
+        let qstr = QString::from(s);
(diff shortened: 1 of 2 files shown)


qttypes/src/qtcore/qstring.rs
```diff
@@ -317,6 +317,12 @@ impl TryFrom<QString> for i16 {
     }
 }
 
+impl From<QString> for Vec<u8> {
+    fn from(x: QString) -> Self {
+        QByteArray::from(x).into()
```
ogoffart, line 322: First you convert from QString to QByteArray,  and then again to to Vec, that's two allocation.
You'd be better of using Rust's DecodeUtf16
(or passing through String instead)

ogoffart (commented):
`AsRef<CStr>` makes sense.  But i'm a bit torn about the other conversion, because they are quite expensive allocation that can, be done otherwise with the existing API.

Ayush1325:
I see, I will check out a better way to achieve this.

PR #266, opened 2022-04-18
Add `iter_mut` and `IndexMut` to `SimpleListItem`

AdrianEddy:
Allow to modify item in place and add a function to notify QML that it has changed


ogoffart:
Sorry for replying so late.

I don't think this is a good idea, because it makes it easy to forget to emit the changed signal.

PR #268, opened 2022-06-17
Expose begin_move_rows and end_move_rows

rubdos:
Sorry for basing this on v0.2.5, I'm still stuck on an old compiler.  Seems like it doesn't conflict, however.


ogoffart:
thanks!

Issue #26, opened 2019-02-15
QColor member results in error.


Pfeil:
I tried to figure out the problem, but somehow I don't get it.
This works: (Minimal example)
```rust
#[derive(QObject, Default)]
struct DataStructure {
    base: qt_base_class!(trait QObject),
    name: qt_property!(QString;),
}
```

But this does not:

```rust
#[derive(QObject, Default)]
struct DataStructure {
    base: qt_base_class!(trait QObject),
    color: qt_property!(QColor;),
}
```

I do not see that/where QString implements QGadget or QObject, so why does it work with QString, but not with QColor? What would I need to change?

Error Message:

```shell
> cargo run
   Compiling qmldemo v0.1.0 (/home/user/Git/qmldemo)
error[E0277]: the trait bound `qmetaobject::QColor: qmetaobject::QGadget` is not satisfied
 --> src/piechart.rs:8:19
  |
8 | #[derive(Default, QObject)]
  |                   ^^^^^^^ the trait `qmetaobject::QGadget` is not implemented for `qmetaobject::QColor`
  |
  = note: required because of the requirements on the impl of `qmetaobject::QMetaType` for `qmetaobject::QColor`
  = note: required because of the requirements on the impl of `qmetaobject::PropertyType` for `qmetaobject::QColor`
  = note: required by `qmetaobject::PropertyType::pass_to_qt`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: Could not compile `qmldemo`.
```

ogoffart:
QColor need to iimplement QMetaType.
The error message is a bit confusing. QGadget do not need to be re-implemented. It could be improved if #[do_not_recommend] was stabilized:   https://github.com/rust-lang/rfcs/pull/2397

ogoffart:
I added the missing QMetaType declaration in commit ec11955

Pfeil:
I see, thanks for that. Now, I have an additional question to that: I have a similar problem with a custom type, I'm not sure whether to open another issue for that. I am currently trying to replicate chapter 4 of this [c++ example][example] where they create a `class PieSlice` which inherits from `QQuickPaintedItem`. They use it as a Q_PROPERTY/qt_property in the `class PieChart`. Since the base member (QObjectCPPWrapper) does/must not implement Clone, I can not derive or simply implement QMetaType for PieSlice (as far as I know).

[...]

PR #270, opened 2022-08-03
Adding an example to illustrate custom QObject nesting

aeyno:
After spending several days figuring out how to put a QObject inside another QObject's struct and finally finding out thanks to [#26](https://github.com/woboq/qmetaobject-rs/issues/26#issuecomment-464503133), I made a little example to illustrate how to do it. I hope it will help other users who are struggling to do something similar.


ogoffart:
Thanks for the patch. 
Sorry for the delay.

PR #272, opened 2022-09-08
Implement QMetaType for QVariantMap

maciek134:
Not sure if there should be any actual tests for this? I based this on https://github.com/woboq/qmetaobject-rs/commit/ec11955458c0ba01b4c25550e182e4c32e4c6c3a


ogoffart:
thanks for the patch.

PR #279, opened 2023-02-18
Fix manipulating UTF-8 strings in QJsonObject

direc85:
During developing Whisperfish, I noticed that pushing emojis into QJsonObject and reading them back didn't work correctly.

Code:
```rust
fn grouped_reactions(&self) -> QByteArray {
    let mut map = std::collections::HashMap::new();

    for (reaction, _) in &self.reaction_list.pinned().borrow().reactions {
        *map.entry(reaction.emoji.clone()).or_insert(0) += 1;
    }
    log::trace!("{:?}", map);
    let mut qmap: qmetaobject::QJsonObject = qmetaobject::QJsonObject::default();
    for (emoji, count) in map {
        qmap.insert(&emoji, QVariant::from(count).into());
    }
    log::trace!("{:?}", qmap.to_json());
    qmap.to_json()
}
```

Output before:

```text
[2023-02-18T19:25:19Z TRACE whisperfish::model::reactions] {"😨": 1}
[2023-02-18T19:25:19Z TRACE whisperfish::model::reactions] {"ð
```

Output after:

```text
[2023-02-18T20:40:38Z TRACE whisperfish::model::reactions] {"😨": 1}
[2023-02-18T20:40:38Z TRACE whisperfish::model::reactions] {"😨":1}
```

The cause was as I suspected: Usage of `QString::fromLatin1(const char *str, qsizetype size)`  doesn't handle UTF-8 encoding correctly, but using `QString::fromUtf8(const char8_t *str, qsizetype size)` is able to handle it.

### DIFF
--- a/qttypes/src/lib.rs
+++ b/qttypes/src/lib.rs
@@ -1770,6 +1770,20 @@ fn test_qjsonobject() {
     assert_eq!(object.keys(), vec!["key".to_owned(), "test".to_owned()]);
 }
 
+#[test]
+fn test_qjsonobject_utf8() {
+    let emoji = String::from("🦀");
+    let expected = String::from("{\"🦀\":1}");
+
+    let mut qmap: QJsonObject = QJsonObject::default();
+    qmap.insert(&emoji, QVariant::from(1).into());
+
+    let actual = qmap.to_json();
+    let actual = actual.to_str().unwrap();
+
+    assert_eq!(actual, expected);
+}
+
 cpp_class!(
     /// Wrapper around [`QJsonArray`][class] class.
     ///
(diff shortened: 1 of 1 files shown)


ogoffart:
Thanks!  the patch looks good.
Could you write a test?

direc85:
First I tried to append the test to the function above, but I couldn't replicate the behavior. I then decided to take the gist of the erroneous code and make a new test out of it.

When I revert the fix, the test fails with
```
---- test_qjsonobject_utf8 stdout ----
thread 'test_qjsonobject_utf8' panicked at 'assertion failed: `(left == right)`
  left: `"{\"ð\u{9f}¦\u{80}\":1}"`,
 right: `"{\"🦀\":1}"`', qttypes/src/lib.rs:1784:5
```

ogoffart:
Thank you!

PR #285, opened 2023-04-09
Add basic QSettings wrapper

direc85:
Upstream-compatible version of https://github.com/whisperfish/qmetaobject-rs/pull/2

---

Attempt of adding QSettings wrapper. This was a bit tricky because QSettings can't be copied, so we have to return `QSettings *` instead, and handle the dereference in the client. This leads into having these in the use-side implementation, but there's no obvious way to get rid of that to me:

```
fn inner(&self) -> &QSettings {
    unsafe { self.inner_ptr.as_ref().unwrap() }
}

fn inner_mut(&mut self) -> &mut QSettings {
    unsafe { self.inner_ptr.as_mut().unwrap() }
}
```


rubdos:
https://github.com/woboq/qmetaobject-rs/actions/runs/4649514331/jobs/8227901989?pr=285 interesting.

direc85:
Interesting indeed. Something about Windows and `tempfile`, I guess. I have a Windows 'puter somewhere still, so I can see what's up with that.

direc85:
It turned out to be a combination of default file extension being `.ini` (on all platforms?) and Linux conveniently using `IniFormat` when set to `NativeFormat`.

I must also say that it would have been convenient to have all tests run to completion even if one test fails. Could that be changed in the future? Thanks...

qttypes/src/qtcore/qsettings.rs
```diff
@@ -0,0 +1,183 @@
...
+    #include <QtCore/QSettings>
+    #include <QtCore/QString>
+}}
+
+cpp_class!(
+    /// Wrapper around [`QSettings`][class] class.
+    ///
+    /// [class]: https://doc.qt.io/qt-5/qsettings.html
+    #[derive(Default)]
+    pub unsafe struct QSettings as "QSettings"
```
ogoffart, line 14: QSettings is a QObject and i'm not confortable having that in a cpp_class that can be moved.

I think i would do  `pub unsafe struct QSettings as "std::unique_ptr<QSettings>"`

direc85:
This just became a blocker to Whisperfish for using master now that we have Rust 1.72 - thanks to the marvelous efforts of @rubdos - so could this please be merged to master? Thanks!

ogoffart:
sorry, i didn't see the updated patch.

direc85:
No worries! Thank you!

PR #288, opened 2023-05-23
Replaced QApplication from QtWidgets by QGuiApplication from QtGui

Atlas42:
When using QML the application class should be a QGuiApplication.
On targets without QtWidgets (my case on an aarch64 with a custom-build qt), the crate does not build. Replacing this class usage fixes it.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,6 +60,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
+        ver: [6.2.0, 6.5.0]
     env:
       QT_QPA_PLATFORM: offscreen
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


ratijas:
IMHO this kind of switch should be configurable. QGuiApplication is more compatible with targets, but QApplication provides automatic style change events and other nice things. How about a cargo feature flag?

Also, your pull request contains lots of unrelated commits, please your change rebase on master.

ratijas:
> Also, your pull request contains lots of unrelated commits, please your change rebase on master.

oh, it's probably because you submitted this pull request from your own master branch, but you continued working on it for your own project. That doesn't work well in practice. Normally, you'd create a new branch per pull request, and only push to it those changes which you intend to land as a part of such pull request.

Atlas42:
Using an other pull request from another branch #294

PR #294, opened 2023-08-17
Feature "qtwidgets" added.

Atlas42:
Link with QtWidgets and use QApplication when activated (default) Use QGuiApplication when not.

Replaces #288 

I rewrote the patch in a new clean branch, keeping the master for my own projects dependency.

As #[cfg] and cpp! does not work well (if not at all together), I used a C++ predefined symbol to configure the used class.
Not as clean as I would like but I lost too much time already to figure out somthing that works.
Tell me if you have a better solution.

### DIFF
--- a/qmetaobject/Cargo.toml
+++ b/qmetaobject/Cargo.toml
@@ -12,9 +12,10 @@ keywords = ["Qt", "QML", "QMetaObject",]
 repository = "https://github.com/woboq/qmetaobject-rs"
 
 [features]
-default = ["log"]
+default = ["log", "qtwidgets"]
 chrono_qdatetime = ["qttypes/chrono"]
 webengine = ["qttypes/qtwebengine"]
+qtwidgets = ["qttypes/qtwidgets"]
 
 [dependencies]
 qttypes = { path = "../qttypes", version = "0.2.0", features = ["qtquick"] }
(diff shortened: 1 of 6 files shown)


ogoffart:
This is a breaking change so it will require a major version update.

Atlas42:
major as in 1.0.0 or as 0.3.0 ?
any idea when that will happen ? not a big fan of keeping custom versions for a long time.

Any idea on implementing this without making it a breaking change ?
I thought bout using a "no-qwidgets" feature instead but I find this rather dirty.

ogoffart:
0.3

Could happen soon.  
@ratijas  is there any other change you would like to do?

ratijas:
nah, it's not like I'm actively working with anything in Rust lately. But thanks for asking!

Maybe we could consider finilizing and merging #227 as well?

ratijas:
Since this PR seems to be replacing #211, how about adding a little note to README just like they did?

ratijas:
> is there any other change you would like to do?

Also, this

> Or we can just keep this QObjectBox as deprecated and remove it when we have reason to break semver.

https://github.com/woboq/qmetaobject-rs/pull/178

Atlas42:
README entry added, my PR is almost identical to #211

Atlas42:
Any news for this or #211 ?

Ciantic:
maybe somebody just hits merge, then we can at least use this via `git` url

ogoffart:
> can't this be done in non breaking fashion?

I don't think so.

PR #300, opened 2023-08-27
Mitigate 128-bit TypeId hash

### LINKED ISSUE #299: Compilation fails with Rust 1.72
```
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> qmetaobject/src/scenegraph.rs:158:32
    |
158 |             raw = cpp!(unsafe [type_id as "quint64", len as "std::size_t"] -> *mut c_void as "QSGNode*" {
    |                                ^^^^^^^
    |
    = note: source type: `TypeId` (128 bits)
    = note: target type: `[u8; 8]` (64 bits)
    = note: this error originates in the macro `__cpp_closure_impl` which comes from the expansion of the macro `cpp` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> qmetaobject/src/scenegraph.rs:163:58
    |
163 |             mask = cpp!(unsafe [raw as "ContainerNode*", type_id as "quint64", len as "std::size_t"] -> u64 as "quint64" {
    |                                                          ^^^^^^^
    |
    = note: source type: `TypeId` (128 bits)
    = note: target type: `[u8; 8]` (64 bits)
    = note: this error originates in the macro `__cpp_closure_impl` which comes from the expansion of the macro `cpp` (in Nightly builds, run with -Z macro-backtrace for more info)
(...)
```

This is caused by Rust 1.72 change [Use 128 bits for TypeId hash](https://github.com/rust-lang/rust/pull/109953/)

direc85:
Obviously, this breaks Rust <1.72, but I couldn't find a neat way to conditionally include the changes using. It doesn't seem to be possible to have something like `#[cfg(version >= "1.72")]` or `#[cfg(TYPE_ID_SIZE == 16)]`, or perhaps I just didn't find the correct keyword to use.

Not sure if either method is good enough for master, but at least there's something that works:

Approach 1: `unsafe { transmute }`

Approach 2: Custom hasher which utilizes `TypeId.hash()` which internally casts `u128` &rarr; `u64`

Improvements and further ideas welcome :sweat_smile:

Fixes #299

### DIFF
--- a/qmetaobject/src/scenegraph.rs
+++ b/qmetaobject/src/scenegraph.rs
@@ -16,8 +16,47 @@ OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 use cpp::cpp;
+use std::{
+    any::TypeId,
+    hash::{Hash, Hasher},
+};
 
 use super::*;
+
+/// Since Rust 1.72, TypeId.t is now u128, so Qt/C++ doesn't accept it.
+/// The field is private, so it can't be simply casted. To prevent unsafe code,
+/// the next best thing is "copy-hasher" which uses TypeId.hash() function,
+/// which does the cast for us. Think of all this as "type_id.t as u64".
+struct TypeIdHasher(u64);
+
+impl Hasher for TypeIdHasher {
+    fn write(&mut self, _: &[u8]) {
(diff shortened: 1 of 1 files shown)


rubdos:
Would it not be better to upsize the older Rust types, instead of chopping the new one?

direc85:
As said in the issue, this is fixed already in master, I'm only affected because I'm running 0.2.5-ish due to Rust 1.52.1. Closing as irrelevant.

ogoffart:
thanks for your patch anyway.

PR #302, opened 2023-10-11
Add a few more QmlEngine methods

IriaSomobu:
I was implementing a basic hot-reload feature for a project of mine and had a difficulties I solved by adding a few additional methods to QmlEngine wrapper.

First, I should briefly explain how hot reloading works in my case:
1. Rust code watches for changes in `./ui/` directroy where all my `.qml`'s are stored;
2. When file change occurs, rust code does the following ([source](https://gitlab.com/somobu/fourteen/-/blob/c05a13dd997e3badd8560c85064cd300447f952b/src/hotreload.rs#L38)):
    - Invokes method `hot_reload_pre` on my root `ApplicationWindow` -- this method [clears](https://gitlab.com/somobu/fourteen/-/blob/c05a13dd997e3badd8560c85064cd300447f952b/ui/Application.qml#L22) `PageLoader`'s `source` and sets it to inactive;
    - Trims and clears QmlEngine's component cache;
    - Invokes method `hot_reload_post` -- this method makes `PageLoader` become active and triggers loading of page where user was before hot reload.

In this scenario I faced two problems:

The first one was caused by cryptic `QMetaMethod::invoke: Unable to invoke methods with return values in queued connections` QmlEngine message happens when I try to call `invoke_method` on engine from my watcher thread. I've looked here and there and found [this answer on SO](https://stackoverflow.com/a/48959406). So I copied `invoke_method`, removed `Q_RETURN_ARG` from a template and renamed method to a `invoke_method_noreturn` (meaning this version of `invoke_method` has no return arg).

The second problem was -- QmlEngine and PageLoader was written by clever guys who cares about performance and do caching if possible. And in my case I need exactly opposite -- I need to reset Engine's/PageLoader's cache on hot reload. I found no possibilities to do it from qml itself (moving my PageLoader to inactive and then back to active state wont help), but I found a way to do it from C side by calling methods `engine->trimComponentCache()` and `engine->clearComponentCache()`. Unfortunately, named methods were absent in wrapper, so I've added them.

And with these two changes I was able to do my simple hot-reload.

I believe someone else may find these addition useful, so here's PR.

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -231,7 +231,8 @@ impl QmlEngine {
         })
     }
 
-    pub fn invoke_method_noreturn(&mut self, name: QByteArray, args: &[QVariant]) -> QVariant {
+    /// This method is the same as [invoke_method] but does not capture or return function's return value
+    pub fn invoke_method_noreturn(&mut self, name: QByteArray, args: &[QVariant]) {
         let args_size = args.len();
         let args_ptr = args.as_ptr();
 
@@ -242,13 +243,12 @@ impl QmlEngine {
             name as "QByteArray",
             args_size as "size_t",
             args_ptr as "QVariant *"
-        ] -> QVariant as "QVariant"
-        {
+        ] {
             auto robjs = self->engine->rootObjects();
             if (robjs.isEmpty()) {
-                return {};
+                return;
             }
-            QVariant ret;
+            
             #define INVOKE_METHOD(...) QMetaObject::invokeMethod(robjs.first(), name __VA_ARGS__);
             switch (args_size) {
                 case 0: INVOKE_METHOD(); break;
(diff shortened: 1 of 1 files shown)


qmetaobject/src/qtdeclarative.rs
```diff
@@ -231,6 +231,54 @@ impl QmlEngine {
         })
     }
 
+    pub fn invoke_method_noreturn(&mut self, name: QByteArray, args: &[QVariant]) -> QVariant {
```
ogoffart, line 234: If the method doesn't return,  why is it returning a `-> QVariant` ?

ogoffart (changes requested):
Thanks for the patch.

The two extra method makes sense.  But the _noreturn one should be changed not to return anything.

qmetaobject/src/qtdeclarative.rs
IriaSomobu: Yup, my bad. I'll fix it a bit later.

By the way `QMetaObject::invokeMethod` itself returns `true` [" if the member could be invoked"](https://doc.qt.io/qt-5/qmetaobject.html#invokeMethod). Maybe it is makes sense to return `Result<QVariant>` in `invoke_method` and `Result<()>` in `invoke_method_noreturn`?

IriaSomobu: Fixed!

As for a `invoke_method` signature change - I think I will open another PR some time later.

ogoffart (approved):
Thanks. 

Tests would be nice if you can add them.

Another thing i'm wondering is whether we should actually call it invoke_method_queued and pass the Qt::QueuedConnection argument.
Is that not more what was your intention?

IriaSomobu:
It is more than I intended indeed -- I just wanted `invoke` to work one way or another, and with this PR it works.

I guess it is possible for me tinker with `invoke_method` and connections a bit later, when there will be more free time.

As for tests -- do you have any ideas on what exactly should we test on these functions?

PR #303, opened 2023-10-14
[WIP] Better invokeMethod

IriaSomobu:
This PR is a follow-up of https://github.com/woboq/qmetaobject-rs/pull/302. This PR introduces breaking changes and not to ready to be merged (yet).

Here I want to discuss and implement changes to make `invoke_method` (and maybe things around it) better.

At first, I've changed `invoke_method`'s return value to `Option<QVariant>` - which is close to original QMetaObject API and makes it possible to check wether error occurred (this [was](https://github.com/woboq/qmetaobject-rs/pull/302#discussion_r1354723902) discussed earlier).

@ogoffart, can you give me advice -- is this change OK as it done now? What should be done next?

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -195,17 +195,25 @@ impl QmlEngine {
         })
     }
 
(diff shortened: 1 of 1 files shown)


IriaSomobu:
Ah yes, the tests

PR #304, opened 2023-11-02
Error out when we detect incompatibilities between mingw and msvc

ogoffart:
This has caused lots of frustrations: eg:
https://github.com/slint-ui/slint/issues/3802
https://github.com/slint-ui/slint/issues/3539
https://github.com/slint-ui/slint/issues/3739

### DIFF
--- a/qttypes/build.rs
+++ b/qttypes/build.rs
@@ -211,6 +211,15 @@ fn main() {
             ""
         };
 
+    // MinGW and MSVC are not compatible
+    if cargo_target_os == "windows" {
+        let spec = qmake_query("QMAKE_SPEC");
+        if (spec.contains("msvc") && cargo_target_env == "gnu")
+            || (spec.contains("g++") && cargo_target_env == "msvc")
+        {
+            report_error(&format!("Rust target '{}' is not compatible with Qt mkspec '{spec}'. Mixing MinGW and MSVC is not allowed.", std::env::var_os("TARGET").unwrap_or_default().to_string_lossy()));
+        }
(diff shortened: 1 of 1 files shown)


qttypes/build.rs
```diff
@@ -211,6 +211,15 @@ fn main() {
             ""
         };
 
+    // MinGW and MSVC are not compatible
+    if cargo_target_os == "windows" {
+        let spec = qmake_query("QMAKE_SPEC");
+        if (spec.contains("msvc") && cargo_target_env == "gnu")
+            || (spec.contains("g++") && cargo_target_env == "msvc")
+        {
+            report_error(&format!("Rust target '{}' is not compatible with Qt mkspec '{spec}'. Mixing MinGW and MSVC is not allowed.", std::env::var_os("TARGET").unwrap_or_default().to_string_lossy()));
```
tronical, line 220: If you want to make it even more useful to users, I suggest to split this into two if statements, each of which can report in `report_error()` what the user can do as remedy, i.e. if spec is mdvc but cargo target env is "gnu": "To fix this, please install Qt compiled for MinGW", and for the other one "To fix this, please install Qt for MSVC".

tronical (approved):

qttypes/build.rs
ogoffart: I believe this is sufficiently clear. They could either solve it by installing a different version of Qt, or getting another build of rust.

(Note that rust users of Slint won't see this error anyway as the warning from the build.rs are hidden.)

PR #305, opened 2023-11-08
Add QQuickWindow and QQuickItem::window_changed

i509VCB:
Some types that I will need to later on bind types related to QSGRendererInterface.

Not related to this pull request, did we want to split `qtdeclarative.rs` up? I suspect that the file will definitely grow quite a bit.

### DIFF
--- a/qmetaobject/src/qtdeclarative.rs
+++ b/qmetaobject/src/qtdeclarative.rs
@@ -248,7 +248,7 @@ impl QmlEngine {
             if (robjs.isEmpty()) {
                 return;
             }
-            
+
             #define INVOKE_METHOD(...) QMetaObject::invokeMethod(robjs.first(), name __VA_ARGS__);
             switch (args_size) {
                 case 0: INVOKE_METHOD(); break;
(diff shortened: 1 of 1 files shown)


qmetaobject/src/qtdeclarative.rs
```diff
@@ -779,6 +779,29 @@ pub fn qml_register_enum<T: QEnum>(
...
+pub struct QQuickWindow(*mut c_void);
+
+impl QQuickWindow {
+    /// # Safety
+    ///
+    /// - The pointer must be a valid `QQuickWindow` pointer.
+    /// - The window must be valid for the specified lifetime `'a`.
+    pub unsafe fn from_raw<'a>(ptr: *mut c_void) -> &'a Self {
+        std::mem::transmute(ptr)
+    }
```
i509VCB, line 796: Now that I realize it, this is undefined behavior...

i509VCB:
Going through dead pull requests, probably won't get to finishing this one due to other priorities.

PR #308, opened 2024-02-22
Qvariant improvements

direc85:
- Add `to_qstring()` wrapper
  - More versatile and efficient than `.to_qbytearray().to_string()`
  - According to [documentation](https://doc.qt.io/archives/qt-5.6/qvariant.html#toString) e.g. `QVariant(QColor)` shouldn't be `toString()`able but it is
- Add `From<QVariantMap>` wrapper
  - Can save a move/clone on Rust side
- Improve debug print
  - Add `type_name` wrapper
  - Include the type QVariant contains with printable value, if any
  - Makes debugging QML<>Rust interactions easier

In Whisperfish for example debug prints like this are now possible:

```rust
tracing::debug!(tracing::debug!("Attachment: {:?}", attachment);
// Attachment: {data: QVariant(QString: "/home/defaultuser/Downloads/test.md"), type: QVariant(QString: "text/markdown")}
```

[Whisperfish!570](https://gitlab.com/whisperfish/whisperfish/-/merge_requests/570) currently depends on this (but it can (barely) do without)

### DIFF
--- a/qttypes/src/qtcore/qvariant.rs
+++ b/qttypes/src/qtcore/qvariant.rs
@@ -75,6 +75,15 @@ impl QVariant {
         })
     }
 
+    /// Wrapper around [`toInt()`][method] method.
+    ///
+    /// [method]: https://doc.qt.io/qt-5/qvariant.html#toInt
+    pub fn to_int(&self) -> u32 {
+        cpp!(unsafe [self as "const QVariant*"] -> u32 as "int" {
+            return self->toInt();
+        })
+    }
+
     /// Wrapper around ['typeName()`][method] method.
     ///
     /// [method]: https://doc.qt.io/qt-5/qvariant.html#typeName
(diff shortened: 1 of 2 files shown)


qttypes/src/qtcore/qvariant.rs
```diff
@@ -57,12 +57,45 @@ impl QVariant {
...
 impl fmt::Debug for QVariant {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(self.to_qbytearray().to_string().as_str())
+        let data = self.to_qstring().to_string();
+        let qtype = self.type_name().to_string();
+        if data.len() == 0 {
+            f.write_fmt(format_args!("QVariant({})", qtype.as_str()))
+        } else {
+            f.write_fmt(format_args!("QVariant({}: \"{}\")", qtype.as_str(), data.as_str()))
+        }
```
rubdos, line 98:
```suggestion
        if data.len() == 0 {
            write!(f, "QVariant({})", qtype.as_str())
        } else {
            write!(f, "QVariant({}: \"{}\")", qtype.as_str(), data.as_str())
        }
```
I think this should work, no?

qttypes/src/qtcore/qvariant.rs
```diff
@@ -57,12 +57,54 @@ impl QVariant {
...
+    }
+
     // FIXME: do more wrappers
 }
 
 impl fmt::Debug for QVariant {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(self.to_qbytearray().to_string().as_str())
+        let data = self.to_qstring().to_string();
+        let qtype = self.type_name().to_string();
```
ogoffart, line 102: it's probably not needed to convert to a rust String here.

PR #309, opened 2024-02-28
Fix feature = "cargo-clippy" deprecation

rex4539:
https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html


ogoffart:
Thanks for the patch!

I think we don't need the cfg_attr anymore and we can just say `#[allow(clippy::..)]`

rex4539:
Let's try again :)

ogoffart:
Thanks

PR #312, opened 2024-08-05
Add null QVariant constructor

direc85:
`QVariant::default()` yields an `undefined` value in QML, but in many places a `null` value is more suited. Add `QVariant::null()` function to achieve that. The test added help distinguishing this.

I would have used `QVariant(nullptr)` instead, but that doesn't seem to work well with the type system in Qt5. The `QJsonValue().toVariant()` is the next best option - and makes sense, since a JSON can also have an explicit `null` value.


ogoffart:
Sorry for not having answer to this PR before.

I don't think it makes sense to construct a QVariant with a JSValue from this API. If you want a JSValue, IMHO you should construct a JSValue and convert that to a QVariant.

direc85:
I have no idea how the test was a merge conflict, but it's now rebased to master.

I agree, it doesn't really make sense to construct a null QVariant from QJsonValue, but that was the most elegant way I was able to find for Qt 5, since `QVariant::fromValue(nullptr)` doesn't work until Qt 6. I mitigated this by making it pick the choice compile time, that's something at least.

But now that I think it twice, and browsed the codebase more, it makes more sense to use `QJsonValue.toVariant()` as it's already available. I think I'll explore that for a bit to see that's usable already.

PR #317, opened 2025-03-22
Add QT bundled libraries behind features

DBLouis:
This patch allows to link QT bundled libraries, behind feature flags. I use this to make fully static builds.

### DIFF
--- a/qttypes/Cargo.toml
+++ b/qttypes/Cargo.toml
@@ -33,6 +33,13 @@ qtsql = []
 # Link against QtTest
 qttest = []
 
+qtbundledfreetype = []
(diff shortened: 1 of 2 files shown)


ratijas:
Needs some public-facing documentation to be viable for merge. Is the developer expected to put these specifically named libs in places?

Nit: The spelling is "Qt", not "QT". Uppercase QT often refers to QuickTime media player by Apple.

DBLouis:
Please tell me where can I write docs for these features.

The idea is that if you need those features, you are compiling Qt yourself, to be linked statically, and those bundled modules are assumed to be built in your Qt install.

PR #320, opened 2025-04-22
Mute clippy transmute warnings

### LINKED ISSUE #291: `QObject` and `QGadget` derives produce clippy warnings
In seemingly all situations, `QGadget` produces a `clippy::transmute_ptr_to_ref` warning.
In a struct with signals, `QObject` produces a `clippy::useless_transmute` for each signal.

Sh3Rm4n:
* `clippy::transmute_ptr_to_ref` for QGadget
* `clippy::useless_transmute` for QObject

Closes #291

### DIFF
--- a/qmetaobject_impl/src/qobject_impl.rs
+++ b/qmetaobject_impl/src/qobject_impl.rs
@@ -686,10 +686,8 @@ pub fn generate(input: TokenStream, is_qobject: bool, qt_version: QtVersion) -> 
     } else {
         quote! {
             #[allow(unused_variables)]
-            let mut obj: &mut #name #ty_generics = &mut *(::std::mem::transmute::<
-                *mut ::std::os::raw::c_void,
-                *mut #name #ty_generics,
-            >(o));
+            #[allow(clippy::transmute_ptr_to_ref)]
+            let mut obj = ::std::mem::transmute::<*mut ::std::os::raw::c_void, &mut #name #ty_generics>(o);
         }
     };
 
@@ -869,7 +867,7 @@ pub fn generate(input: TokenStream, is_qobject: bool, qt_version: QtVersion) -> 
                 // FIXME!  we should probably use the signature verbatim
                 let n = &arg.name;
                 let ty = &arg.typ;
-                quote! { mut #n : #ty }
+                quote! { #n : #ty }
             })
             .collect();
         let args_ptr: Vec<_> = signal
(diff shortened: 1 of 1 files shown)


qmetaobject_impl/src/qobject_impl.rs
```diff
@@ -875,7 +875,12 @@ pub fn generate(input: TokenStream, is_qobject: bool, qt_version: QtVersion) ->
             .map(|arg| {
                 let n = &arg.name;
                 let ty = &arg.typ;
-                quote! { unsafe { ::std::mem::transmute::<& #ty, *mut ::std::os::raw::c_void>(& #n) } }
+                quote! {
+                    {
+                        let mut n = #n;
+                        (&mut n as *mut #ty as *mut ::std::os::raw::c_void)
```
ogoffart, line 881: that doesn't look right. It seems it's taking the address of `n` on the stack.

qmetaobject_impl/src/qobject_impl.rs
```diff
@@ -686,7 +686,7 @@ pub fn generate(input: TokenStream, is_qobject: bool, qt_version: QtVersion) ->
     } else {
         quote! {
             #[allow(unused_variables)]
-            let mut obj = ::std::mem::transmute::<*mut ::std::os::raw::c_void, &mut #name #ty_generics>(o);
+            let mut obj = &*o;
```
ogoffart, line 689: i would think you still need to cast to the right type explicitly.

ogoffart (commented):
thanks for the PR.

But i'm not sure if this is correct, see inline comments.

qmetaobject_impl/src/qobject_impl.rs
Sh3Rm4n: I'm not sure either. This should be catched by the compiler though. The code I've tested did not generate any errors. The `let` binding was necessary to get a mutable reference `&mut`. The compiler suggests to annotate the function argument with `mut`. I'll check if that is possible.

ogoffart: In this case it might not be catched by the compiler since the user of the pointer uses unsafe code.

qmetaobject_impl/src/qobject_impl.rs
Sh3Rm4n: Yeah this solution looks wrong. I kept the transmute to cast to the right type as `*mut`, but dereferenced it now to `&mut` outside of the transmute.

qmetaobject_impl/src/qobject_impl.rs
Sh3Rm4n: Good point. I now use marked the function argument as `mut` so that a `&mut` reference can be taken.

Sh3Rm4n:
Overall these changes are all inside of `unsafe` blocks, and I'm not to confidident, if I got the semantics right without invoking UB. The alternative would be to just mute the clippy warnings with `[allow(...)`. WDYT @ogoffart?

Sh3Rm4n:
Alright, I'm not sure what caused this failure <https://github.com/woboq/qmetaobject-rs/actions/runs/14596502246/job/40950448403>, I could reproduce it locally, but also after reverting my changes (if I haven't done anything incorrectly). 

As my goal is to mute the clippy warnigs, let's see if `[allow(...)]` makes the test run again.

ogoffart:
I think this could be done without transmute, only with pointer cast

Sh3Rm4n:
Alright, with the last CI run https://github.com/woboq/qmetaobject-rs/actions/runs/14612927424/job/41009015057 and the current change https://github.com/woboq/qmetaobject-rs/compare/master...Sh3Rm4n:qmetaobject-rs:4622085296b20e2fb27bf081d64c9216a1502c04 (which does no code changes) I think the CI failure is not related to this PR

PR #328, opened 2026-03-26
offset_off! instead of unsafe offset computation

rubdos:
@ogoffart 

> Could the Rust's offset_of be used?

The answer is yes! Had no idea this existed.


ogoffart (approved):
Great, thanks

PR #330, opened 2026-04-12
chore: bump syn to 2

rubdos:
The whole ecosystem seems to have moved on to syn 2, including cpp/cpp_build etc. These dependencies don't see that as a breaking change. AFAICT, neither should we, because all the parse trees stay internal.

I took the liberty of flattening the nesting of the parts this change touched.


ogoffart (approved):
thanks

